PR #11, opened 2018-08-07
configure: fix with-network default case

shalzz:
Using the --with-network switch without an argument
as suggested in the install instructions,
has a default value of "yes" set by automake.
Consider that as a valid value for the
"testnet" case as well which is our intended default.


chjj:
Good call. Merging.

Issue #5, opened 2018-04-14
Heap corruption during authoritative resolution


chjj:
Spotted this today when clicking an obfuscated twitter link. The daemon had been running locally on my laptop for a few days.

```
rs: query
rs:   id=27745
rs:   labels=2
rs:   name=t.co.
rs:   type=1
rs:   class=1
rs:   edns=0
rs:   dnssec=0
rs:   tld=co
rs:   addr=127.0.0.1:51557
rs: udp nodata
ns: query
ns:   id=14389
ns:   labels=2
ns:   name=t.co.
ns:   type=1
ns:   class=1
ns:   edns=1
ns:   dnssec=1
ns:   tld=co
ns:   addr=127.0.0.1:52340
ns: udp nodata
corrupted double-linked list (not small)
Aborted
```

I've been unable to reproduce it, so no chance of using valgrind to track this down. The `ns: udp nodata` log implies that `hsk_ns_onrecv()` successfully executed otherwise I would suspect this of being an issue with the authoritative cache that was just added. It's also not sending a message in response, so there's no cache hit there.

My best guess is something funky happened in the P2P pool which corrupted the heap. I'm starting to add more debug logs so we can narrow this down when it happens again.

chjj:
Update: Finally a lead after so many months of head scratching...

When implementing the [unbound] node module, I noticed a similar heap corruption in node.js. It only presented itself when the unbound context was set to async mode. It was consistent and reproducible (it seemed to cause unrecoverable memory corruption maybe 1 out of every 20 times the [bns] test suite was run). I'm going to guess the same issue is affecting hnsd.

I hesitate to call this a bug in libunbound. It's possible that libuv and libevent don't play well with each other for some reason (?).

I think the solution for now would be to call unbound's resolver synchronously in the uv [thread pool][pool] (the same fix used in the unbound node module).

We can leave this open to investigate the causes of this more thoroughly in the future.

[unbound]: https://github.com/chjj/unbound
[bns]: https://github.com/chjj/bns
[pool]: http://docs.libuv.org/en/v1.x/threadpool.html

pinheadmz:
This has come up again in a branch where hnsd can discover peers and open more connections: https://github.com/handshake-org/hnsd/pull/38#issuecomment-681883190

pinheadmz:
Got a stack trace of this:
```
peer 714 (64.227.15.172:12038): sending verack                                                                                                                                                    [25/1965]
[...]

PR #14, opened 2019-05-10
Windows support

JonathonH-LTM:
Support native Windows builds with Msys2/MinGW, using the MinGW libunbound package.

The major issue was that libunbound can't be hooked up to the libuv event loop with a file descriptor like on Unix (libunbound does not use a pipe on Windows, and libuv on Windows can only poll sockets anyway).
Instead, create a worker thread to operate libunbound, using libuv's threading primitives.  Results are queued back to the libuv event loop using a libuv async.
hnsd now shuts down cleanly on SIGINT and SIGTERM too, realized it was not doing this when I could not test the thread shutdown code.  This uncovered minor issues in the pool shutdown code, which are fixed.
The libunbound path can also be set when configuring, we're building libunbound from source in order to stay on the same version on all platforms.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -19,7 +19,8 @@ CLEANFILES =
 
 lib_LTLIBRARIES = libhsk.la
 
-libhsk_la_LIBADD = $(top_builddir)/uv/libuv.la
+libhsk_la_LIBADD = $(top_builddir)/uv/libuv.la   \
+                   $(LIB_WINBCRYPT)
 
 libhsk_la_CFLAGS = -DHSK_BUILD @CFLAGS@
 libhsk_la_LDFLAGS = -no-undefined -version-info 0:0:0
--- a/Makefile.am
+++ b/Makefile.am
@@ -67,7 +67,8 @@ hnsd_SOURCES = src/cache.c  \
                src/daemon.c \
                src/ns.c     \
                src/rs.c     \
-               src/rs_worker.c
+               src/rs_worker.c \
+               src/signals.c
 
 hnsd_LDADD = $(LIB_UNBOUND)             \
              $(top_builddir)/libhsk.la
--- a/Makefile.am
+++ b/Makefile.am
@@ -67,11 +67,11 @@ hnsd_SOURCES = src/cache.c  \
                src/ns.c     \
                src/rs.c
 
-hnsd_LDADD = -lunbound                  \
+hnsd_LDADD = $(LIB_UNBOUND)             \
              $(top_builddir)/libhsk.la
 
 hnsd_LDFLAGS = -static
-hnsd_CFLAGS = -DHSK_BUILD $(AM_CFLAGS)
+hnsd_CFLAGS = -DHSK_BUILD $(INC_UNBOUND) $(AM_CFLAGS)
 hnsd_CPPFLAGS = $(AM_CPPFLAGS)
 
 # pkgconfigdir = $(libdir)/pkgconfig
--- a/Makefile.am
+++ b/Makefile.am
@@ -66,7 +66,8 @@ noinst_PROGRAMS = $(PROGS)
 hnsd_SOURCES = src/cache.c  \
                src/daemon.c \
                src/ns.c     \
-               src/rs.c
+               src/rs.c     \
+               src/rs_worker.c
 
 hnsd_LDADD = $(LIB_UNBOUND)             \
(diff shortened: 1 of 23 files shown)


JonathonH-LTM:
This snowballed a little bit, so let me know if you'd rather I break this up and make separate PRs.

Thanks!

tynes:
Do you think you could add sections to the `README` regarding Windows Installation and Dependencies? I'm in the process of testing this and I'm figuring out how to build it on Windows, it would make it way easier for me to run it

JonathonH-LTM:
Absolutely, I'll add that and push it up.

JonathonH-LTM:
Let me know if there are any dependency issues after installing per those notes, I kept notes as I was setting it up but it's possible that I could have missed something.

tynes:
From a fresh Windows development environment, I was able to build this PR on Windows following the instructions. It was able to sync to the tip and resolve DNS queries using `dig`. I then built this PR on Linux just to make sure that still worked, and I was able to sync to the tip and resolve DNS queries using `dig` again.

chjj:
Whoa, great work. That's interesting that windows was having some event loop issues. I was considering dropping the async resolution altogether and just doing synchronous resolutions on the libuv thread pool (this sounds stupid, I know, but I have a lingering suspicion that uv+unbound were causing a heap corruption when [used together][1]).

I will play around with this tomorrow. I only wish I had a windows machine to test this on.

[1]: https://github.com/handshake-org/hnsd/issues/5#issuecomment-461303114

JonathonH-LTM:
Thanks Chris, looking forward to hearing how this works for you after you try it.  I've been running it on Linux and Mac too.

I'd hesitate to block the libuv event loop with any synchronous recursive resolutions.  We've found that on connections with relatively high latency (starting around >150 ms or so), uncached resolution times can be very long, often several seconds.  AFAIK this is inherent to running a recursive resolver locally, latency costs you a lot for all the round trips.  The cache helps mitigate this, but many domains we've tested are only cached for ~5 minutes.

I'll keep an eye out for the possible heap corruption.  I can't think of a way libuv and libunbound would interact poorly to cause that, they don't really directly interact very much (even less with this PR), though anything's possible of course 😛 

I did notice in Valgrind that we get some invalid reads/writes on hnsd shutdown.  It looks like these are mostly due to freeing libuv resources improperly - uv_close() is async, so the object can't be freed until the callback is invoked.  Some uv objects seem to be freed right away, like hsk_rs_t.socket; I used after_close_free() in this PR to free resources after they're closed.  These only occur on shutdown though in hnsd, so it probably doesn't explain any heap corruption.

JonathonH-LTM:
I noticed that libunbound is causing the worker thread to busy-wait at 100% CPU - apparently ub_wait() doesn't actually block if there are no requests pending.  I'm working on this, will probably have the fix up Monday.

JonathonH-LTM:
Busy-waiting is fixed, shutdown of the worker thread is a bit cleaner now too.
Also ended up fixing the libuv resources in rs/ns/pool that weren't beeing freed properly, was doing some shutdown validation with Valgrind and they were causing a lot of errors.

JonathonH-LTM:
@chjj Have you had a chance to check this out?

tynes:
I vote to make @JonathonH-LTM a maintainer of this repo with commit access, if he wants it that is. Its important to not let the ecosystem fragment too much with many stale PRs and various forks. Newcomers may end up using a vulnerable fork which would lead to a less secure network.

JonathonH-LTM:
Thanks for the vote of confidence Mark, but for now I think I'd rather not take on that additional responsibility.  Even with commit access it'd still mean a lot to have JJ's seal of approval 😁.  I don't anticipate many more changes from our side for the foreseeable future though.

I appreciate you guys taking the time to review these changes 👍.  I want to avoid making this a long-term fork too if possible, and I might revisit commit access if we did start doing more work on this down the road.

JonathonH-LTM:
@boymanjor Can we merge this?  Would be a big help if we don't have to carry these patches downstream, and Windows support is on the TODO list for hnsd anyway.

boymanjor:
@JonathonH-LTM I'm going to merge this now, but it is worth noting that hnsd does not currently work with the hsd network protocol. @chjj has mentioned a possible, complete rewrite of hnsd. I've halted work on hnsd until we hear more from him, but we might as well get this in now.

PR #18, opened 2019-09-17
etc: add hnsd systemd service

tuxcanfly:
WIP: looking for feedback to improve defaults, conflicts with other resolvers, tests on various linux distros etc.

### DIFF
--- a/etc/hnsd.service
+++ b/etc/hnsd.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=handshake DNS resolver
+Wants=network-online.target
+After=network-online.target
+
+[Service]
+EnvironmentFile=/etc/hnsd.conf
+ExecStart=/usr/bin/hnsd $POOL_SIZE $RS_HOST $UNBOUND $PEER
+AmbientCapabilities=CAP_NET_BIND_SERVICE
+Restart=always
+RestartSec=3
+
+[Install]
+WantedBy=default.target
--- a/etc/hnsd.conf
+++ b/etc/hnsd.conf
@@ -1,0 +1,3 @@
+POOL_SIZE=--pool-size=1
+RS_HOST=--rs-host=127.0.0.1:53
+UNBOUND=--rs-config=/etc/unbound/unbound.conf


etc/hnsd.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=handshake DNS resolver
+Wants=network-online.target
+After=network-online.target
+
+[Service]
+EnvironmentFile=/etc/hnsd.conf
+ExecStart=/usr/bin/hnsd $POOL_SIZE $RS_HOST $UNBOUND $PEER
```
pinheadmz: Does `$PEER` need a value in `etc/hnsd.conf`? Also, why use env variables like this instead of just writing the arguments in the `ExecStart` line?

tuxcanfly: iirc you can't have arguments for the `ExecStart` line, thus the need for this hack. I agree if there's a better way to handle this we should probably use that.

pinheadmz: I think you can, at least it works on Ubuntu. Here's my service file for the full node at https://easyhandshake.com

```
[Unit]
Description=hsd
After=network.target

[Service]
Type=simple
Restart=always
RestartSec=1
User=root
ExecStart=/root/n/bin/node /root/hsd/bin/hsd --no-wallet

[Install]
WantedBy=multi-user.target
```

tuxcanfly: OK I found that you can't add args if the args have a space in them. So... we can point `hnsd` to `/etc/hnsd.conf` so it will be easily configurable.

tuxcanfly:
So I don't think the config file is actually being parsed. Instead what I've done is pass the config options as command line params through the use of `EnvironmentFile` directive in systemd service.

tynes:
Since `hnsd.conf` is no longer being parsed as a config file, should it be renamed to something like `hnsd.env` and have each of the possible environment variables in it but commented out? We would also need to add the environment variable parsing to `hnsd` for this to work.

etc/hnsd.service
```diff
@@ -0,0 +1,38 @@
...
+After=systemd-sysusers.service systemd-networkd.service
+Before=network.target nss-lookup.target shutdown.target
+Conflicts=shutdown.target
+Wants=nss-lookup.target
+
+[Service]
+AmbientCapabilities=CAP_SETPCAP CAP_NET_RAW CAP_NET_BIND_SERVICE
+CapabilityBoundingSet=CAP_SETPCAP CAP_NET_RAW CAP_NET_BIND_SERVICE
+EnvironmentFile=/etc/hnsd.conf
+ExecStart=/usr/bin/hnsd $HNSD_CONFIG
```
tynes: Were you planning on removing `$HNSD_CONFIG`?

tuxcanfly: No, its injected from `/etc/hnsd.conf`. It's a bit of hack to resemble a configuration file, but all it does is set the env var that gets passed to the executable.

This is only necessary since we don't have config parsing as of yet.

faddat: I'd just recommend that flags be used and a sane default be put in the service file.

eg:

hnsd -r 127.0.0.1:53

PR #20, opened 2019-09-18
dns: fix loop in hsk_dns_label_decode_tlsa

jacobhaven:
`hsk_dns_label_decode_tlsa` was never decoding the port properly. In fact, it the port's first character was `'0'`, it would enter an infinite loop.

### DIFF
--- a/src/dns.c
+++ b/src/dns.c
@@ -2800,6 +2800,8 @@ hsk_dns_label_decode_tlsa(const char *name, char *protocol, uint16_t *port) {
 
     if (word > 0xffff)
       return false;
+
+    s += 1;
   }
 
   if (port)


boymanjor:
LGTM. Nice catch!

boymanjor:
@jacobhaven slight nit re: the description - I believe the code path results in an infinite loop _unless_ the first character in the port is `\0`, right?

jacobhaven:
@boymanjor No, @chjj's style is a bit confusing here but the ASCII digit `'0'` = `0x30`, so this loop bails out if the char is not (inclusively) between `'0'` and `'9'`. It also bails out if the port it's reading is above `0xffff` (which seems reasonable).

But, since it never advances to the next char, it reads `"443"` as `444444` and bails out. It label it tries to read starts with a `'0'` (e.g. `"0123"`) it will read `000000...` and infinitely loop.

In any case, it never reads the port string correctly without advancing the loop.

boymanjor:
Ah, I see. All valid ASCII digit prefixes, save `0`, will eventually trip the 2-byte limit check. `000000...` never increases in value.

PR #24, opened 2019-11-26
Remove cuckoo

tynes:
This PR removes dead code in this repo that involves the old PoW, Cuckoo Cycle. These files are no longer used:

- `src/cuckoo.c`
- `src/cuckoo.h`

Also remove Cuckoo Cycle related constants and the Cuckoo Cycle related error messages. Made sure to make sure the remaining errors match up regarding their integer to string mapping.


kilpatty:
LGTM!

src/error.h
```diff
@@ -26,37 +26,27 @@
...
-#define HSK_EPOWTOOSMALL 21
-#define HSK_EPOWNONMATCHING 22
-#define HSK_EPOWBRANCH 23
-#define HSK_EPOWDEADEND 24
-#define HSK_EPOWSHORTCYCLE 25
-
 // POW
-#define HSK_ENEGTARGET 26
-#define HSK_EHIGHHASH 27
+#define HSK_ENEGTARGET 19
```
kilpatty: Just curious, are there any side effects for changing the error numbers here?

tynes: The integers defined in this file match up with the error message strings in `src/error.c`, see https://github.com/handshake-org/hnsd/pull/24/files#diff-ea12193fb130aea2da140db9c638871bR45

PR #25, opened 2019-12-12
genesis: regenerate

tynes:
Regenerate the genesis block after https://github.com/handshake-org/hsd/commit/a1a2231a60654c6375699b4ce9c2120c25d723bf

See: https://github.com/handshake-org/hsd/pull/307


boymanjor:
@tynes this genesis block has gone stale. I'm going to close this for now.

PR #29, opened 2020-05-18
Mainnet 1/4: Update constants

pinheadmz:
Sets all mainnet constants including genesis block, network magic bytes, port numbers and ICANN root zone "snapshot".

Adds new constant `HSK_BRONTIDE_PORT` and sets it for all the default port usage in `pool`.

Imports seed nodes with identity keys from hsd. TODO: implement better peer discovery with brontide-enabled peers, OR implement clearnet networking in hnsd. At this point, the 8 seed nodes are the only peers hnsd will connect to by default (additional nodes can be added by starting with `-s pubkey@IP:port`).

References:

https://github.com/handshake-org/hsd/commit/a5e282a57fd99fcfe6bbd82f02894193a9514643

https://github.com/handshake-org/hsd/commit/62e413a0cee26d4546cc8100ac5f645830d42d21

https://github.com/handshake-org/hs-names/commit/2e4df74f2df04db80b2a7082ee06fcbc8129e5aa


configure.ac
```diff
@@ -76,14 +76,14 @@ AC_ARG_WITH([network],
     [Specify network. Default is testnet.]
```
kilpatty, line 76: I think this should be switched to "Default is main/mainnet"

kilpatty:
Looking good to me :)

configure.ac
boymanjor: Agreed. I think we should use "main." It matches the expected command line argument.

pinheadmz: thanks, fixed.

pinheadmz:
rebase to  35e231024eefd37979fe281f897eff5bd11a12f4:

Fix nit from review comment.
Update `tld.h` with latest icann root zone file as of today, some updates: 


```diff
@@ -40,7 +40,6 @@ static const char *HSK_TLD_NAMES[] = {
   "agency",
   "ai",
   "aig",
-  "aigo",
   "airbus",
   "airforce",
   "airtel",
@@ -55,6 +54,7 @@ static const char *HSK_TLD_NAMES[] = {
   "alsace",
   "alstom",
   "am",
+  "amazon",
   "americanexpress",
   "americanfamily",
   "amex",
@@ -399,7 +399,6 @@ static const char *HSK_TLD_NAMES[] = {
   "es",
   "esq",
   "estate",
-  "esurance",
   "et",
   "etisalat",
   "eu",
@@ -845,7 +844,6 @@ static const char *HSK_TLD_NAMES[] = {
   "mz",
   "na",
   "nab",
-  "nadex",
   "nagoya",
   "name",
   "nationwide",
@@ -1028,7 +1026,6 @@ static const char *HSK_TLD_NAMES[] = {
   "rich",
   "richardli",
   "ricoh",
-  "rightathome",
   "ril",
   "rio",
   "rip",
@@ -1079,7 +1076,6 @@ static const char *HSK_TLD_NAMES[] = {
   "schwarz",
   "science",
   "scjohnson",
-  "scor",
   "scot",
   "sd",
   "se",
@@ -1178,7 +1174,6 @@ static const char *HSK_TLD_NAMES[] = {
   "sx",
   "sy",
   "sydney",
-  "symantec",
   "systems",
   "sz",
   "tab",
@@ -1291,7 +1286,6 @@ static const char *HSK_TLD_NAMES[] = {
   "virgin",
   "visa",
   "vision",
-  "vistaprint",
   "viva",
   "vivo",
   "vlaanderen",
@@ -1385,6 +1379,7 @@ static const char *HSK_TLD_NAMES[] = {
   "xn--c1avg",
   "xn--c2br7g",
   "xn--cck2b3b",
+  "xn--cckwcxetd",
   "xn--cg4bki",
   "xn--clchc0ea0b2g2a9gcd",
   "xn--czr694b",
@@ -1395,7 +1390,6 @@ static const char *HSK_TLD_NAMES[] = {
   "xn--e1a4c",
   "xn--eckvdtc9d",
   "xn--efvy88h",
-  "xn--estv75g",
   "xn--fct429k",
   "xn--fhbei",
   "xn--fiq228c5hs",
@@ -1421,12 +1415,12 @@ static const char *HSK_TLD_NAMES[] = {
   "xn--j1aef",
   "xn--j1amh",
   "xn--j6w193g",
+  "xn--jlq480n2rg",
   "xn--jlq61u9w7b",
   "xn--jvr189m",
   "xn--kcrx77d1x4a",
   "xn--kprw13d",
   "xn--kpry57d",
-  "xn--kpu716f",
   "xn--kput3i",
   "xn--l1acc",
   "xn--lgbbat1ad8j",
@@ -1467,7 +1461,6 @@ static const char *HSK_TLD_NAMES[] = {
   "xn--otu796d",
   "xn--p1acf",
   "xn--p1ai",
-  "xn--pbt977c",
   "xn--pgbs0dh",
   "xn--pssy2u",
   "xn--q7ce6a",
```

dnsguru:
Good to see, but would be better to augment this with : https://www.icann.org/resources/registries/gtlds/v2/gtlds.json which will indicate gTLD deltas (not ccTLD or IDNccTLD)

pinheadmz:
> Good to see, but would be better to augment this with : https://www.icann.org/resources/registries/gtlds/v2/gtlds.json which will indicate gTLD deltas (not ccTLD or IDNccTLD)

Aren't all the gTLD's included in the root zone? https://www.internic.net/domain/root.zone

dnsguru:
No, that is not going to contain some of the gtld names that are contracted
but not in the root, plus it can be an indicator of where brands might
cancel their TLD.

This is how the project got the .music/Amazon IDN/.spa/.kids et al issue

pinheadmz:
If their not in the root, then how do they resolve? This update is specifically for resolving DNS from ICANN, it has nothing to do with reserved names or anything like that, just DNS.

dnsguru:
We use this https://github.com/tld-update-bot to keep the
https://github.com/publicsuffix/list  crispy and in alignment.  We update
frequently, and have different purpose, but if there's anything there which
is helpful, by all means why not use it.

I realize the projects and needs are distinct, but I think there is a
factor to consider, which is the span of time between pulls on the HSD
project maybe missing some things that get introduced, and this would allow
for proactive inclusion (and removals) to better address names that would
be in the DNS between updates of the HSD.

When the gTLD thundering herd came along in 2013-> names were getting added
and some software companies that 'think different' would incorporate the
PSL into their browsers on their desktops and mobile/tablet OS and
resolvers, which would cause odd routing until the next update of their OS
which would pull in deltas.  So every x months they'd incorporate the list
and it would fix 'broken' TLDs from going to search instead of DNS.

The json file adds names once they enter into contract with ICANN, which
can sometimes precede the introduction to the root by enough lead time to
allow for the gap time between updates.

PR #30, opened 2020-05-18
Mainnet 2/4: Elligator

pinheadmz:
Adds Elligator Squared pubkey obfuscation to "Act One" of the Brontide handshake. Crypto code is imported from bcrypto and exposed to the brontide module. 

References:

https://github.com/handshake-org/hsd/commit/e0a2f7b35ccbd9e7406db14b7fa3756941dbd4e3

### DIFF
--- a/src/brontide.c
+++ b/src/brontide.c
@@ -25,7 +25,7 @@
 #include "utils.h"
 
 static const char brontide_protocol_name[] =
-  "Noise_XK_secp256k1_ChaChaPoly_SHA256";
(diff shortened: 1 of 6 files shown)


kilpatty:
Since brontide w/ Elligator deprecates the version entirely you can probably kill this as well: https://github.com/handshake-org/hnsd/blob/master/src/brontide.c#L35

kilpatty:
Finally got to the rest of this today! LGTM

PR #31, opened 2020-05-18
Mainnet 3/4: Fix VERACK response in p2p handshake

pinheadmz:
hnsd was sending its `getheaders` message in response to a `VERACK` from the peer. This is bad timing as the handshake is not yet complete, and the hsd node will disregard the headers request. I'm not sure how this managed to work during testnet, except that perhaps there was actually a race condition with the messages in transit, or maybe the new Elligator code is messing with the timing.

Either way, without this patch mainnet still seems to connect but regtest fails immediately due to the incomplete p2p handshake.

References:

https://github.com/handshake-org/hsd/blob/1d7251957dd5ce35eb173c3fcdca0ca2479dcb83/lib/net/peer.js#L474-L523

### DIFF
--- a/src/pool.c
+++ b/src/pool.c
@@ -1143,6 +1143,7 @@ hsk_peer_send(hsk_peer_t *peer, const hsk_msg_t *msg) {
 
 static int
 hsk_peer_send_version(hsk_peer_t *peer) {
+  hsk_peer_log(peer, "sending version\n");
   hsk_pool_t *pool = (hsk_pool_t *)peer->pool;
 
   hsk_version_msg_t msg = { .cmd = HSK_MSG_VERSION };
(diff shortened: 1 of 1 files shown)


kilpatty:
LGTM

src/pool.c
```diff
@@ -1143,6 +1143,7 @@ hsk_peer_send(hsk_peer_t *peer, const hsk_msg_t *msg) {
 
 static int
 hsk_peer_send_version(hsk_peer_t *peer) {
+  hsk_peer_log(peer, "sending version\n");
```
tuxcanfly, line 1146: Nit: maybe we use `hsk_debug_log`?

pinheadmz: I think `hsk_peer_log()` is correct, all the other p2p message logs in this file use it, and it is peer specific, eg:

```
peer 4 (172.104.214.189:44806): connected
peer 4 (172.104.214.189:44806): sending version
peer 4 (172.104.214.189:44806): received verack
peer 4 (172.104.214.189:44806): received version: /hsd:2.0.2/ (28839)
peer 4 (172.104.214.189:44806): sending verack
peer 4 (172.104.214.189:44806): sending sendheaders
peer 4 (172.104.214.189:44806): sending getheaders
peer 4 (172.104.214.189:44806): received 1 addrs
```

PR #32, opened 2020-05-18
Mainnet 4/4: Resource serialization

pinheadmz:
Update resource serialization to mainnet version 0 as implemented in hsd. Most record types were removed from what we allow on the blockchain. All thats left is `DS`, `NS`, `GLUE`, `SYNTH`, and `TXT`. Of these, `DS` records changed very little (if at all?) and so I followed the patterns around that as far as memory usage and flow through the program.

Other details about the change:

- An extra string compressor was removed

- NS records don't have an IP address

- TXT records actually contain an array of strings

I assume this hsd PR is correct in its intent around SYNTH and GLUE records:

https://github.com/handshake-org/hsd/pull/444

(If the hsd PR is incorrect then this hnsd PR will not be incompatible, but behavior between hnsd and hsd resolvers will differ)

TODO: Add a testing suite to this repo. For development, my process included running [this script](https://gist.github.com/pinheadmz/ca0e1a5739607307c7a515158ad89c98) against a local hsd node in regtest, and then using dig to retrieve the records from hnsd:

`dig @127.0.0.1 -p 25349 test-ns NS`

etc.

References:

https://github.com/handshake-org/hsd/issues/125

https://github.com/handshake-org/hsd/commit/a72bab62a019e97488f8d5248dbb983649532823

https://github.com/handshake-org/hsd/commit/68e55773d745d1caacc6ef512d3607dfc4a4ec57


pinheadmz:
Latest commit cleans up GLUE when the NS name is not a subdomain of the queried name. See https://github.com/handshake-org/hsd/pull/444#issuecomment-634003174

src/resource.c
```diff
@@ -56,997 +39,410 @@ ip_write(const uint8_t *ip, uint8_t *data);
...
 
-  if (!hsk_resource_host_read(data, data_len, st, &rec->target))
-    return false;
+    // Size of this string
+    uint8_t size = 0;
+    if (!read_u8(data, data_len, &size)) {
+      hsk_dns_txt_free(txt);
+      return false;
+    }
+
```
tuxcanfly, line 157: nit: there's an extra whitepsace somewhere here.

pinheadmz: thanks,  cleaned up

PR #42, opened 2020-09-12
respond with servfail if bogus

buffrr:
When validating DNSSEC, the resolver should return SERVFAIL if DNSSEC validation fails (bogus = true) 

example:

`dig @1.1.1.1 dnssec-failed.org a`

will return SERVFAIL

hnsd only omits the ad flag from the response which makes downgrade attacks possible.

~~Note: hnsd also considers handshake tlds without a DS RR bogus (they are not secure but they are also not bogus). This should also be fixed because this PR will break handshake tlds that don't use DNSSEC.~~

Fixed by adding an NSEC record to prove that a DS doesn't exist which allows unbound to treat it as unsigned instead of bogus.

### DIFF
--- a/src/resource.c
+++ b/src/resource.c
@@ -23,6 +23,11 @@ static const uint8_t hsk_type_map[] = {
   0x00, 0x00, 0x03, 0x80
 };
 
+// NS RRSIG NSEC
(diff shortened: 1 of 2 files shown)


pinheadmz:
Closed by #76 lets move discussion / review there.

Issue #28, opened 2020-04-23
Fails build on MacOS 10.15.4


kdesigns:
When following the build process on my mac it stops on an error:

`configure: error: could not determine ar interface`

Full output here:

```
+ pushd uv
~/hnsd/uv ~/hnsd
+ test -f configure
+ ./autogen.sh
+ glibtoolize --copy
glibtoolize: putting auxiliary files in '.'.
glibtoolize: copying file './ltmain.sh'
glibtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'm4'.
glibtoolize: copying file 'm4/libtool.m4'
glibtoolize: copying file 'm4/ltoptions.m4'
glibtoolize: copying file 'm4/ltsugar.m4'
glibtoolize: copying file 'm4/ltversion.m4'
glibtoolize: copying file 'm4/lt~obsolete.m4'
+ aclocal -I m4
+ autoconf
+ automake --add-missing --copy
configure.ac:38: installing './ar-lib'
configure.ac:25: installing './compile'
configure.ac:22: installing './config.guess'
configure.ac:22: installing './config.sub'
configure.ac:21: installing './install-sh'
configure.ac:21: installing './missing'
Makefile.am: installing './depcomp'
+ popd
~/hnsd
+ glibtoolize --copy
glibtoolize: putting auxiliary files in '.'.
glibtoolize: copying file './ltmain.sh'
glibtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'm4'.
glibtoolize: copying file 'm4/libtool.m4'
glibtoolize: copying file 'm4/ltoptions.m4'
glibtoolize: copying file 'm4/ltsugar.m4'
glibtoolize: copying file 'm4/ltversion.m4'
glibtoolize: copying file 'm4/lt~obsolete.m4'
+ aclocal -I m4
+ autoconf
+ automake --add-missing --copy
configure.ac:12: installing './ar-lib'
configure.ac:10: installing './compile'
configure.ac:6: installing './config.guess'
configure.ac:6: installing './config.sub'
configure.ac:4: installing './install-sh'
configure.ac:4: installing './missing'
Makefile.am: installing './depcomp'
checking for a BSD-compatible install... /usr/local/bin/ginstall -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /usr/local/bin/gmkdir -p
checking for gawk... no
checking for mawk... no
checking for nawk... no
checking for awk... awk
[...]

PR #47, opened 2020-11-12
Github actions

faddat:
This adds github actions for Mac and Linux.

Mac is currently not working due to #28.   

I'm working on Mac, and I think I should be able to get it resolved fairly quickly.

### DIFF
--- a/.github/workflows/hnsd.yml
+++ b/.github/workflows/hnsd.yml
@@ -27,6 +27,21 @@ jobs:
       run: ./configure
     - name: make
       run: make
+  windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: msys2/setup-msys2@v2
+        with:
+          msystem: MSYS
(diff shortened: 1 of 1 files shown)


faddat:
Just need to -Syyu the windows one, then restart the build after.

pinheadmz:
Thanks for getting this going. @faddat -- would you be able to:

- rebase on master now that https://github.com/handshake-org/hnsd/pull/48 is merged
- squash these commits into a single commit

Hopefully we can see the action run successfully on your fork of the repo and if it works we can merge here.

We can address the Windows build error after that, once we verify the OSX fix.

[Github actions docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions) state that `macos-latest` is 10.15 so that should include whatever update to Clang 12 causes the implicit declaration to be an error.

pinheadmz:
Closed by https://github.com/handshake-org/hnsd/pull/68. Thank you!

PR #48, opened 2020-11-13
fixed implicit declaration by importing elligator.h to ec.h

faddat:
An implicit declaration was blocking builds on macs with clang 12 and creatinng warnings on all platforms.  

This resolves that issue by importing elligator.h in ec.h  ﻿


faddat:
Closes #28

PR #50, opened 2021-01-08
add Dockerfile and Readme

magicstone1412:


pinheadmz:
I don't use Docker regularly, would be good to get a review on this perhaps from @blusyn or @Thann

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+RUN apk add git automake autoconf libtool unbound-dev
+
+COPY . /
+
+RUN ./autogen.sh && ./configure && make
+
+FROM alpine:latest
+
+RUN apk update && apk upgrade
```
Thann, line 14: Typically I:
1. call the first chuck "BASE"
2. Do an apk "update --no-cache [runtime deps]"
3. Then I do "from base as builder" so the discarded building phase uses the updates from the base
4. Then the final chunck is "from base" and already has the runtime deps

This make the overall build as fast and small as possible.
 =]

Thann (commented):
It looks pretty good to me, I only had one minor suggestion so speed up build time. Also there could be some security hardening I'll look into that now =]

magicstone1412:
> @magicstone1412 can you find a reviewer for this PR who has some public github activity involving docker expertise?

I don't know who can review my PR. Anyway, @Thann reviewed already.

I read a book to write the Dockerfile. This is the first project I contribute. Therefore, please help me correct it until you think it's okay.

Thann:
I think adding this would be best for security, although it may break things for windows users

`RUN useradd -ms /bin/bash hnsduser`
`USER hnsduser`

pinheadmz:
@magicstone1412 thanks for contributing! sorry it took so long to review. I'm gonna hijack this PR over here: https://github.com/handshake-org/hnsd/pull/58 and try to get it finished today.

PR #55, opened 2021-02-09
Update req.h

sebastianrasor:
missing * means hnsd won't compile on openbsd

### DIFF
--- a/src/req.h
+++ b/src/req.h
@@ -31,7 +31,7 @@ typedef struct {
   char tld[HSK_DNS_MAX_LABEL + 1];
 
   // Who it's from.
-  struct sockaddr_storage ss;
+  struct sockaddr_storage *ss;
   struct sockaddr *addr;
 } hsk_dns_req_t;
 


pinheadmz:
Man building on openbsd is hard! Finally installed everything I needed and had to run configure like this:

`./configure --with-unbound=/usr/local/ CC=egcc`


But then I was able to reproduce your error and confirm this branch fixes it:

```

**Comment by pinheadmz** [comment]
Hm although this now:

```
assertion "sa && dst" failed: file "src/addr.c", line 564, function "hsk_sa_to_string"
Abort trap (core dumped)
```

sebastianrasor:
Yeah I wasn’t able to get past that

pinheadmz:
Ok we obviously need to fix that before merging. My guess without digging in is that the original code is correct (it's not a pointer to a struct) and the "incomplete type" error means something else. Do you have any spare cycles to keep working on this? If not we can leave this PR open, I'll come back around to it after addressing some other issues.

PR #60, opened 2021-03-16
pool: make unknown command debug instead of log

tuxcanfly:
`Unknown command` in the logs looks a bit spammy. Make it debug instead.

### DIFF
--- a/src/pool.c
+++ b/src/pool.c
@@ -1675,7 +1675,7 @@ hsk_peer_parse(hsk_peer_t *peer, const uint8_t *msg, size_t msg_len) {
   const char *str = hsk_msg_str(peer->msg_cmd);
 
   if (strcmp(str, "unknown") == 0) {
-    hsk_peer_log(peer, "unknown command: %u\n", peer->msg_cmd);
+    hsk_peer_debug(peer, "unknown command: %u\n", peer->msg_cmd);
     goto done;
   }
 


pinheadmz:
Oh clever -- how do you set `HSK_DEBUG_LOG` ? Do you pass it to `configure` or have to manually add it to makefile?

nodar-chkuaselidze:
You can pass to configure:
 `CFLAGS=-DHSK_DEBUG_LOG ./configure` (or `./configure CFLAGS=-DHSK_DEBUG_LOG`)
and `libhsk` will have `HSK_DEBUG_LOG` defined.

Maybe there's another way as well ?

pinheadmz:
@nodar-chkuaselidze thanks ok cool. Do you guys think we should add "configure options" section to readme? The other one I use all the time is `configure --with-network=regtest` for development. And when I was installing on openBSD I had to pass `--with-unbound=/usr/local/`

nodar-chkuaselidze:
It can definitely be useful in some form, 

e.g. HSK_DEBUG_LOG may not be useful for local builds but can be useful for development and debugging - but this option is kinda obscure.
As for `configure` options, personally I don't think it's necessary. `configure --help` is good enough to me.

Maybe document commonly used options ? (e.g. with-network - even though you can see it in configure help)

nodech:
I think it is good idea to make it DEBUG.

pinheadmz:
@nodech I think we can avoid this altogether by merging https://github.com/handshake-org/hnsd/pull/63

pinheadmz:
https://github.com/handshake-org/hnsd/pull/63 has been merged. I don't think we should merge this PR at this time, but i dunno if we need to close it. If some other message type starts spamming hnsd logs, users will report it and we should look in to it.

pinheadmz:
Closing this for now

PR #61, opened 2021-03-16
chain: add header store

tuxcanfly:
This PR adds a persistent header store to the chain. The headers are read from the disk on startup, and processed as usual. This should be fairly safe as chain work is validated every time. The store is backed by a mmaped file (currently written to `/tmp/hnsd.bin`). The format is pretty straightforward, it's just the header encoded to bytes and written to disk.

Some work remaining here:

* Write number of headers written somewhere (first byte?) so that one can allocate memory accordingly.
* Page the mmap file after first 65536 headers are processed (currently this is hardcoded and will break after block 65536).
* Multi-OS testing (mmap doesn't work on windows so maybe switch to normal file).

### DIFF
--- a/src/chain.c
+++ b/src/chain.c
@@ -112,6 +112,8 @@ hsk_chain_init_genesis(hsk_chain_t *chain) {
   chain->tip = tip;
   chain->genesis = tip;
 
+  hsk_store_write(chain->store, data, 236);
+
   hsk_chain_maybe_sync(chain);
 
   return HSK_SUCCESS;
--- a/src/chain.c
+++ b/src/chain.c
@@ -730,6 +730,7 @@ hsk_chain_insert(
   uint8_t *data = malloc(236);
   hsk_header_encode(hdr, data);
   hsk_store_write(chain->store, data, sizeof(data));
+  free(data);
 
   return HSK_SUCCESS;
 }
--- a/src/chain.c
+++ b/src/chain.c
@@ -89,11 +89,19 @@ hsk_chain_init(hsk_chain_t *chain, const hsk_timedata_t *td, const uv_loop_t *lo
   chain->store = hsk_store_alloc(chain->loop);
   hsk_store_open(chain->store);
 
-  for (int i = 0; i < 236; i++) {
+  int rc = hsk_chain_init_genesis(chain);
+  if (rc != HSK_SUCCESS) {
+    return rc;
+  }
+
+  // TODO: hardcoded 65536 because we don't know how many headers we've written
+  // So we read until we hit a header which has garbage data and break
+  // Here we're checking if bits == 0
+  for (int i = 1; i < 65536; i++) {
     hsk_header_t *hdr = hsk_header_alloc();
-    hsk_store_read(chain->store, 65536, hdr);
+    hsk_store_read(chain->store, i, hdr);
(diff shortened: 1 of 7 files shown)


src/store.c
```diff
@@ -0,0 +1,153 @@
...
+static void
+hsk_store_timer(hsk_store_t *store) {
+  hsk_store_sync(store);
+}
+
+int
+hsk_store_open(hsk_store_t *store) {
+  // TODO: hardcoded for 65536 blocks. Should auto re-size if we see more than
+  // 65536 blocks.
+  store->size = 65536 * 236;
```
nodar-chkuaselidze, line 55: Is this related to mmap limits or this is just temporary number until proper fix?

src/chain.c
```diff
@@ -57,23 +58,73 @@ qsort_cmp(const void *a, const void *b) {
...
+  for (int i = 1; i < 65536; i++) {
+    hsk_header_t *hdr = hsk_header_alloc();
+
+    if (!hdr)
+      return HSK_ENOMEM;
+
+    hsk_store_read(chain->store, i, hdr);
+
+    // FIXME: ugly hack ugh!
+    if (hdr->bits == 0) {
```
nodar-chkuaselidze, line 116: Does this come from `ftruncate` expanding file with `0`s ?

tuxcanfly: Yep, this is to check if we're seeking above the size we've written. If we're reading a null byte block, because we're going above the written size, we will parse a header with bits = 0 (among other fields).

I think the one way to fix this hack is to create a file with size = number of headers * sizeof header, then we can os.Stat on it to get the exact number of headers we need to be seeking to.

src/store.c
tuxcanfly: This a temporary number. It _will_ break after block 65536 :cry:

src/store.c
```diff
@@ -0,0 +1,153 @@
...
+#include <fcntl.h>
+#include <unistd.h>
+
+#include "utils.h"
+#include "error.h"
+#include "header.h"
+#include "store.h"
+#include "errno.h"
+
+#define HSK_HEADERS_BIN "/tmp/hnsd.bin";
```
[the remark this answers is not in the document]
tuxcanfly: Yes, I think we can use `uv_os_tmpdir` from libuv since we're already depending on it.

src/store.c
```diff
@@ -0,0 +1,153 @@
...
+    return HSK_ENOMEM;
+  store->timer = NULL;
+  return HSK_SUCCESS;
+}
+
+int hsk_store_uninit(hsk_store_t *store) {
+  store->map = NULL;
+  store->location = NULL;
+  store->headers = NULL;
+  if (munmap(store->map, store->size) == -1)
```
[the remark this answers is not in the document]
tuxcanfly: I think we have to rethink the approach of using memory-mapped files. Originally I thought that since we handle re-orgs in memory, if we're mapping the headers from memory to the mmap file, we don't have to re-implement re-org when writing to the file. I'm not so sure if this works like I had planned. Also, mmap doesn't seem to work on windows and I'm not sure about BSD.

The alternative is to just use a normal file.

pinheadmz:
closing for https://github.com/handshake-org/hnsd/pull/99

PR #66, opened 2021-07-27
pool: check for errors in after_read before using the buffer

buffrr:
This PR fixes a frequent crash on windows whenever a connection resets with error -4077. This crash should technically happen on other OSs but I've only noticed it frequently on windows.

```
Thread 1 received signal SIGSEGV, Segmentation fault.
0x00007ffb1ceb43cf in msvcrt!memmove () from C:\Windows\System32\msvcrt.dll
(gdb) bt
#0  0x00007ffb1ceb43cf in msvcrt!memmove ()
   from C:\Windows\System32\msvcrt.dll
#1  0x00007ff6bf897801 in hsk_peer_on_read (peer=0x1b68af72ee0,
    data=<optimized out>, data_len=<optimized out>) at src/pool.c:1603
#2  0x00007ff6bf898206 in hsk_peer_on_read (data_len=18446744073709547539,
    data=<optimized out>, peer=0x1b68af72ee0) at src/pool.c:1595
#3  after_read (stream=<optimized out>, nread=-4077, buf=<optimized out>)
    at src/pool.c:1818
#4  0x00007ff6bf8d0f51 in uv_process_tcp_read_req (
    loop=loop@entry=0x7ff6bf972120 <default_loop_struct>,
    handle=0x1b68af72ef8, req=req@entry=0x1b68af72f78) at src/win/tcp.c:976
#5  0x00007ff6bf8be64b in uv_process_reqs (
    loop=0x7ff6bf972120 <default_loop_struct>) at src/win/req-inl.h:159
#6  uv_run (loop=0x7ff6bf972120 <default_loop_struct>, mode=UV_RUN_DEFAULT)
    at src/win/core.c:503
#7  0x00007ff6bf883367 in main (argc=1, argv=0x1b688541970)
    at src/daemon.c:621
```

the `after_read` function [here](https://github.com/handshake-org/hnsd/blob/master/src/pool.c#L1800) is getting called with nread=-4077 a negative nread value means [an error]( http://docs.libuv.org/en/v1.x/stream.html#c.uv_read_cb) in this case it's [ECONNRESET](https://github.com/handshake-org/hnsd/blob/master/uv/include/uv-errno.h#L134)

~~Also, on error hnsd only destroys the peer but it doesn't free data in the buffer like in this [example](https://github.com/handshake-org/hnsd/blob/master/uv/test/dns-server.c#L243)  so it could leak a bit of memory whenever an error occurs and the buffer is not empty~~

```
The callee is responsible for freeing the buffer, libuv does not reuse it. The buffer may be a null buffer 
(where buf->base == NULL and buf->len == 0) on error.
```

### DIFF
--- a/src/pool.c
+++ b/src/pool.c
@@ -1802,7 +1802,14 @@ after_read(uv_stream_t *stream, ssize_t nread, const uv_buf_t *buf) {
 
   if (!peer)
     return;
-
+  
+  if (nread < 0) {
+    if (nread != UV_EOF)
+      hsk_peer_log(peer, "read error: %s\n", uv_strerror(nread));
+    hsk_peer_destroy(peer);
+    return;
+  }
+  
   if (peer->brontide != NULL) {
     int r = hsk_brontide_on_read(
         peer->brontide,
@@ -1820,13 +1827,6 @@ after_read(uv_stream_t *stream, ssize_t nread, const uv_buf_t *buf) {
         (uint8_t *)buf->base,
         (size_t)nread
         );
-  }
-
-  if (nread < 0) {
-    if (nread != UV_EOF)
-      hsk_peer_log(peer, "read error: %s\n", uv_strerror(nread));
-    hsk_peer_destroy(peer);
-    return;
   }
 }
 


src/pool.c
```diff
@@ -1802,7 +1802,17 @@ after_read(uv_stream_t *stream, ssize_t nread, const uv_buf_t *buf) {
 
   if (!peer)
     return;
+  
+  if (nread < 0) {
+    if (buf->base)
+      free(buf->base);
```
pinheadmz, line 1808: I'm not sure if this is necessary. `buf->base` is pointed to a preallocated buffer in the peer struct:

https://github.com/handshake-org/hnsd/blob/c4d0a9ffd823cd605c7ba8358982157071d64137/src/pool.c#L1795

...which is allocated:
https://github.com/handshake-org/hnsd/blob/c4d0a9ffd823cd605c7ba8358982157071d64137/src/pool.h#L64

...and initialized when each new peer is born:

https://github.com/handshake-org/hnsd/blob/c4d0a9ffd823cd605c7ba8358982157071d64137/src/pool.c#L818

...and then the entire `peer` struct is freed when we destory the peer, which we're about to do here in 4 lines:

https://github.com/handshake-org/hnsd/blob/c4d0a9ffd823cd605c7ba8358982157071d64137/src/pool.c#L891

pinheadmz (approved):
Reproduced issue on windows using deviously modified full node, this patch prevents the segfault.

ACK 3145f7f25c3842e0c1b01a90224f91d5452d764c

```
Hash: SHA256

ACK 3145f7f25c3842e0c1b01a90224f91d5452d764c
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmECytQACgkQ5+KYS2KJ
yTojAg//TA7re4g1kpTIyZf4eEZaN9NDTTmcf1f4lFsj4veJvgfyS3pFJmhFkq6R
71p7LrrVDUFWQbKXakgqnhXXvfguvFPxUbLjxTiqOgyfWE56yMeWjbhna137DlWD
1D+f8zniTF4vgRCzTvuNK253pGuZkuEYc+DTG1hyunfsMII+41pqTJ4TVmrJHwui
bhT5Z39WUmuwu7HiHFd9odnEe6wR+a2Cqyxe4RcTC4pH7zqMNqIMITh8SzMzu5tF
Kt8kXK/rb+W1MKiyzK0gRO5FWpJkqcl1MxkrtYtU4e4JS/9WIFQb+aVuZOps43Kv
8hi8BnjAn6C1MbplU00hWgjBYy+Ub9iN8rfn6hk/bSpnfPFI9Zdnx1rJwHowhBzH
vm6p6VCvlGUwIs2UpOSsP411Ac03RlWcHaPIfpCGXN7EVOaxq6dJ0sx4KhSz7mSE
3eNBXie0IxoaY+1U4RqTDr48sFvqj/Df94X6/NFUOePBnHquMWCV60SJ/TdQMWOg
K7WuaCjfdaGtb40qvSshSjrQL2+Hrz27TqeqZzlfmAEPiXO6fF03FBqwdH9kOcI3
ANZ7964SYeMWkcdAFYPLoK/ocmUBdIoZeTimQwuw3bq9U43H8AA0GfxcEBBa8Pqt
u0FFKBXL40FoXkvjGQ+pWgwKhEbrTOUNtiJZCx3TQDgdvJzoUeU=
=bVXx
-----END PGP SIGNATURE-----
```

</details>

PR #67, opened 2021-07-31
ns/resource: fix SYNTH records

pinheadmz:
Way back in https://github.com/handshake-org/hnsd/pull/32 I made a few mistakes that left SYNTH records broken in hnsd. While they could be _returned_ from a resource if the root ns was queried directly for an HNS name with a SYNTH record, a few things did not work:

 - root ns would decode and answer queries for synth records on the fly
 - when generating glue from a SYNTH record, the hostname was incorrect (left in initial state)

On mainnet HNS I have a name `omnitude` with both SYNTH4 and SYNTH6 records, and a nameserver and webserver in production at the corresponding IP. For review and testing, I ran hsd in regtest and just "won" that name again so I could insert the mainnet record set:

```
{
  "records": [
    {
      "type": "SYNTH4",
      "address": "45.77.219.32"
    },
    {
      "type": "DS",
      "keyTag": 26614,
      "algorithm": 8,
      "digestType": 2,
      "digest": "a20bc6f9ada0883326a05374d0d0c0e6290cef580d00b9b95770301441733b7f"
    },
    {
      "type": "SYNTH6",
      "address": "2001:19f0:5:450c:5400:3ff:fe31:2ccc"
    }
  ]
}
```

Then you can `./configure --with-network=regtest` and `make` hnsd with and without this PR to see the problem and its solution:

## Master 

glue records missing hostname from root ns:

```
--> dig @127.0.0.1 -p 25349 omnitude ns

...

;; AUTHORITY SECTION:
omnitude.               21600   IN      NS      _1bvvubadrcg0._synth.
omnitude.               21600   IN      NS      _84g026fg0l2gol000fvvsc9cpg._synth.

;; ADDITIONAL SECTION:
.                       21600   IN      A       45.77.219.32
.                       21600   IN      AAAA    2001:19f0:5:450c:5400:3ff:fe31:2ccc
```

root ns not decoding and answering synth records:

```
--> dig @127.0.0.1 -p 25349 _1bvvubadrcg0._synth.

;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 43875
;; flags: qr aa rd; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1
;; WARNING: recursion requested but not available

...

;; AUTHORITY SECTION:
.                       86400   IN      SOA     . . 2021073116 1800 900 604800 86400
```

recursion failure on name with synth record:

```
--> dig @127.0.0.1 -p 25350 omnitude

[...]

PR #68, opened 2021-08-06
Introduce unit tests

pinheadmz:
This commit was cherry-picked from #67 so we can flesh it out and get it working on windows (apparently only working on OSX and linux right now).

### DIFF
--- a/.github/workflows/test-windows.yaml
+++ b/.github/workflows/test-windows.yaml
@@ -1,0 +1,34 @@
+name: Test - Windows
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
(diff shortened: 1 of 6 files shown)


pinheadmz:
CI and test results can be seen on this mirrored PR to m personal master with github actions already enabled: https://github.com/pinheadmz/hnsd/pull/2

buffrr (approved):
This is a great and simple way to add initial tests/CI to hnsd.

It would be cool if we update this later to support the standard `make check` command by autotools which also runs the tests by libuv. We can update this format once we have a lot more tests & figure out a way to run only specific tests/or at least exclude libuv tests because they take a long time to run

PR #70, opened 2021-08-25
ns: fix caching and not-caching synth records

pinheadmz:
Fixes a bug in https://github.com/handshake-org/hnsd/pull/67 where I re-used a variable name `synth` for both a string and a flag indicating whether or not we should cache a response.


buffrr (approved):
The `synth` definition was shadowed in the inner scope but technically it shouldn't affect the definition in the outer scope. Either way this makes it more readable. LGTM.

PR #72, opened 2021-09-19
Makefile for OpenWRT

myorn:


Anunayj:
I think we should start to version hnsd better to avoid confusion with which version corresponds to what package in build systems in future? and only update the makefile when there is a major version change.

Currently the v1.0.0 is tagged as 462220f761e967d753fc5a16a31aace9c9c7a6ef

Oh and there is no gpg signature on the tags, which would be nice to have.

Anunayj:
Was able to build the hnsd package locally with the SDK, and was also able to browse DANE+handshake using letsdane+hnsd running on a OpenWRT router. 

image
image

Anunayj (approved):
Just need to update the package version and add a gpg signed tag for the latest commit.

I'm a little confused about how many package architectures are there though? What's the difference between `aarch64_armv8-a` v/s `aarch64_cortex-a53` v/s `aarch64_cortex-a72`.

Would a single binary not work on all of these?

openwrt_Makefile
```diff
@@ -0,0 +1,53 @@
...
+PKG_INSTALL:=1
+PKG_BUILD_PARALLEL:=1
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/hnsd
+  SECTION:=net
+  CATEGORY:=Network 
+  TITLE:=hnsd
+  URL:=https://hnsd.org/
```
Anunayj, line 35: Change the URL to something else? Does anyone own this domain?

myorn: Oh, I'm so sorry, I totally forgot about it. I'll fix it in a moment

myorn:
> I'm a little confused about how many package architectures are there though? What's the difference between `aarch64_armv8-a` v/s `aarch64_cortex-a53` v/s `aarch64_cortex-a72`.
> 
> Would a single binary not work on all of these?

We had two devices, one with cortex-a53 and the other with cortex-a72. Packages from one didn't work with the other. 
I'm not really sure why, and I really want to know.

openwrt_Makefile
pinheadmz: should be https://handshake.org

openwrt_Makefile
```diff
@@ -0,0 +1,53 @@
...
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=hnsd
+PKG_VERSION:=1
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/handshake-org/hnsd.git
+PKG_SOURCE_VERSION:=c60b2b49adf6de92f028670f1608493dbbad1578
+PKG_MIRROR_HASH:=d2990a59bf091dda43a939ef76c5733be1fc1e727783c27b488fc49908e165ad
```
pinheadmz, line 18: I see `PKG_SOURCE_VERSION` is the latest commit hash -- whats `PKG_MIRROR_HASH`? Could you also add a comment about how to update these for future releases? We actually just merged an important memory leak fix and this makefile should already be updated for that.

README.md
```diff
@@ -283,6 +283,21 @@ To stop a container named `hnsd`, run:
...
+and put it to `your_openwrt_dir/package/net/hnsd` before building.
+Then you can use your `menuconfig` and select it.  
+Or you can use this command if you want to build on your SDK this package only:
+
+```bash
+$ make package/net/hnsd/compile V=s
+```
+
+Please keep in mind that `hnsd` needs `libunbound` and all of its dependencies
+such as `libsodium, libmnl, libevent2(all packs), libpthread, libnghttp2, python3-base,libprotobuf-c`
```
pinheadmz, line 298: Is this as easy as just
`opkg install libsodium libmnl,libevent2 libpthread libnghttp2 python3-base libprotobuf-c` ?

myorn: Almost like it. I would suggest to try to enter them one by one and troubleshoot. For the most part of the supported architectures it should be easy (maybe for any, I just don't have a possibility to check them all). OpenWRT sometimes just not seeing that some packages are supported by default and you'll have to install them from the lowest dependencies to the top.

openwrt_Makefile
myorn: > whats PKG_MIRROR_HASH?

> PKG_MIRROR_HASH is also SHA256 and is used when you download the source with git protocol based on commit hash. As buildbot places the download sources to the server as .tar.xz, the next build can use the PKG_MIRROR_HASH to verify the package without calling upstream git server, and can re-use the source file instead of re-downloading it with git. (In practice you first once let the Make process to download the .tar.xz to your dl directory and then you can use that .tazr.xz to calculate the hash an place it to the Makefile.)

> Could you also add a comment about how to update these for future releases?
Sure, should I do it in the `openwrt_Makefile` also?
https://openwrt.org/docs/guide-developer/packages#use_source_repository
It's here, main part is this one:
image

Also, there's tag variable `PKG_MAINTAINER` that can be specified in the `openwrt_Makefile` and I wanted to ask, do you want me to leave my email there? It's not a big problem for me to maintain it but I'm not available all the time unfortunately :)

myorn: > We actually just merged an important memory leak fix and this makefile should already be updated for that.

Which version should I put in the `PKG_VERSION` then?

Issue #45, opened 2020-09-24
Memory leak


pinheadmz:
Running master branch at 95da9e3556adb5f5996d6070fb57ba76a0615afa

Command: `sudo valgrind -s --track-origins=yes --leak-check=full ./hnsd -r 127.0.0.1:53`
Run for a few hours, normal system DNS along with two threads of [this](https://github.com/handshake-org/hnsd/issues/43)

valgrind output:

```
==262390== 
==262390== HEAP SUMMARY:
==262390==     in use at exit: 4,111,477 bytes in 16,868 blocks
==262390==   total heap usage: 3,638,404 allocs, 3,621,536 frees, 1,501,481,665 bytes allocated
==262390== 
==262390== 192 bytes in 3 blocks are definitely lost in loss record 5 of 21
==262390==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==262390==    by 0x4C3759D: __gmp_default_allocate (in /usr/lib/x86_64-linux-gnu/libgmp.so.10.4.0)
==262390==    by 0x4BC9FE7: _nettle_gmp_alloc_limbs (in /usr/lib/x86_64-linux-gnu/libhogweed.so.5.0)
==262390==    by 0x4BCD9B8: nettle_ecc_point_init (in /usr/lib/x86_64-linux-gnu/libhogweed.so.5.0)
==262390==    by 0x48D3591: ??? (in /usr/lib/x86_64-linux-gnu/libunbound.so.8.1.4)
==262390==    by 0x48D3A6F: ??? (in /usr/lib/x86_64-linux-gnu/libunbound.so.8.1.4)
==262390==    by 0x48D5099: ??? (in /usr/lib/x86_64-linux-gnu/libunbound.so.8.1.4)
==262390==    by 0x48D5E77: ??? (in /usr/lib/x86_64-linux-gnu/libunbound.so.8.1.4)
==262390==    by 0x48D6565: ??? (in /usr/lib/x86_64-linux-gnu/libunbound.so.8.1.4)
==262390==    by 0x48D7576: ??? (in /usr/lib/x86_64-linux-gnu/libunbound.so.8.1.4)
==262390==    by 0x48D7B97: ??? (in /usr/lib/x86_64-linux-gnu/libunbound.so.8.1.4)
[...]

PR #73, opened 2021-09-23
Cache memleak

nodech:
This is next memleak found with `make CFLAGS=-fsanitize=leak` and fixes another
leak logged in https://github.com/handshake-org/hnsd/issues/45 by @pinheadmz.

It's highly likely to be related to this:
```
==262390== 73,597 bytes in 294 blocks are definitely lost in loss record 18 of 21
==262390==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==262390==    by 0x12B315: hsk_dns_msg_encode (dns.c:178)
==262390==    by 0x120CC3: hsk_cache_insert (in /home/pinheadmz/Desktop/work/hnsd/hnsd)
==262390==    by 0x1222E8: after_resolve (in /home/pinheadmz/Desktop/work/hnsd/hnsd)
```

You can replicate it with sanitizer this way:
  - `make CFLAGS=-fsanitize=leak`
  - `./hnsd -r 127.0.0.1:1553`
  - wait for some time to sync (I just wait til 10k)
  - `dig @127.0.0.1 -p 1553 handshake.org`
  - wait for it to resolve (sometimes it may with different error in unbound,
    this PR is not about that)
   - C-c hnsd

You should see following message:
```
=================================================================
==663041==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 11595 byte(s) in 8 object(s) allocated from:
    #0 0x7f987da5abc2 in __interceptor_malloc /build/gcc/src/gcc/libsanitizer/lsan/lsan_interceptors.cpp:56
    #1 0x55d3f0524371 in hsk_dns_msg_encode src/dns.c:178

SUMMARY: LeakSanitizer: 11595 byte(s) leaked in 8 allocation(s).
```

This PR is an attempt to fix that leak.

Story of the `hsk_dns_msg_encode`:
  `hsk_dns_msg_encode` is responsible to create `wire` representation of the
`hsk_dns_msg_t` struct. It is *allocates* buffer for the wire representation
and leaves responsibility of freeing that data to the caller.

  Major callers for this method are: `rs.c`, `ns.c` via `hsk_dns_msg_finalize`
method(defined in `req.c`) and `cache.c` when inserting into cache in
`hsk_cache_insert`. `rs.c` and `ns.c` are similar in the structure and both
use `wire` with `hsk_rs_send` and `hsk_ns_send` which both free this data
after using (as far as I have seen) - you can see `should_free` variable and
`after_send` callbacks in work.

  `hsk_cache_insert` on the other hand does not free this buffer in all cases.
After this fix new behaviour of `hsk_cache_insert_data` will be: if it is
successful it will free the data (only when cache is still valid) and if it
returns false it will become `hsk_cache_insert` methods responsibility to free
that data.

This could be avoided if we had more strict rules about the allocation/free
responsibilities and could be different (and better) solution to this:

  Move allocation of the wire buffer outside of the `hsk_dns_msg_encode`, we
would call `hsk_dns_msg_size` from the outside everywhere we need `wire`
serialization to allocate.
It would require slightly more refactor and decided to keep it minimal with
this PR.

### DIFF
--- a/src/cache.c
+++ b/src/cache.c
@@ -81,8 +81,10 @@ hsk_cache_insert_data(
   hsk_cache_item_t *cache = hsk_map_get(&c->map, &ck);
 
   if (cache) {
-    if (hsk_now() < cache->time + 6 * 60 * 60)
+    if (hsk_now() < cache->time + 6 * 60 * 60) {
+      free(wire);
       return true;
+    }
 
     hsk_map_del(&c->map, &ck);
     hsk_cache_item_free(cache);
@@ -105,7 +107,8 @@ hsk_cache_insert_data(
   item->time = hsk_now();
 
   if (!hsk_map_set(&c->map, &item->key, item)) {
-    free(item->msg);
+    // hsk_cache_insert will free msg on false
+    item->msg = NULL;
     free(item);
     return false;
   }
@@ -129,6 +132,7 @@ hsk_cache_insert(
 
   if (!hsk_cache_insert_data(c, req->name, req->type, wire, wire_len)) {
     hsk_cache_log(c, "could not insert cache\n");
+    free(wire);
     return false;
   }
 


src/cache.c
```diff
@@ -129,6 +132,7 @@ hsk_cache_insert(
 
   if (!hsk_cache_insert_data(c, req->name, req->type, wire, wire_len)) {
     hsk_cache_log(c, "could not insert cache\n");
+    free(wire);
```
pinheadmz, line 135: 👍  had to double check that `hsk_dns_msg_encode` would not return `false` if the malloc was successful (otherwise I'd expect a `free()` in the previous block as well).

pinheadmz:
Code LGTM but I'm having trouble executing the modified code path (I inserted a bunch of printfs) I can verify the leak on master branch but on the PR branch I haven't been able to hit those lines to make sure the leak is plugged.

What's the logic?

The root ns has to try to insert into cache something that is already there? That has been there less than 6 hours? I don't know why I can't use `dig` to make that happen reliably.

src/cache.c
nodech: yeah, this makes code pretty unreadable.
That's the reason I mention to refactor so the callers should allocate the wire.

nodech: But because this malloced data later is stored as reference in the cache map, so we can't free it
also make it harder to follow the logic.

pinheadmz:
Ok I figured out how to reliably prove this fixes the leak. 

First, never cache anything:

```diff
diff --git a/src/ns.c b/src/ns.c
index 1325c59..a37e949 100644
--- a/src/ns.c
+++ b/src/ns.c
@@ -311,7 +311,7 @@ hsk_ns_onrecv(
   hsk_dns_msg_t *msg = NULL;

   // Hit cache first.
-  msg = hsk_cache_get(&ns->cache, req);
+  msg = false;

   if (msg) {
     if (!hsk_dns_msg_finalize(&msg, req, ns->ec, ns->key, &wire, &wire_len)) {
```

...this ensures that `hsk_cache_insert_data()` gets called on every request.

Then start hnsd (I built for regtest for faster iteration) and tie to another local hsd peer:

```
 MallocStackLogging=1 ./hnsd -s 127.0.0.1:10000
```

Then simply request the same name from the root NS twice in a row:

```
 dig @127.0.0.1 -p 25349 com
 dig @127.0.0.1 -p 25349 com
```

And get the leak data:

```
--> ps | grep hnsd
32336 ttys012    0:00.01 grep hnsd
32333 ttys015    0:00.06 ./hnsd -s 127.0.0.1:10000
--> leaks 32333
...
```

The result here is bloody!


```
STACK OF 1 INSTANCE OF 'ROOT LEAK: malloc<1536>':
9   libdyld.dylib                      0x7fff203f9f3d start + 1
8   hnsd                                  0x1021ffd38 main + 1896
7   hnsd                                  0x102289261 uv_run + 449  core.c:369
6   hnsd                                  0x1022989ef uv__io_poll + 1199  kqueue.c:0
5   hnsd                                  0x1022913c6 uv__stream_io + 1974  stream.c:1316
4   hnsd                                  0x102226f93 hsk_peer_on_read + 2403  pool.c:1620
3   hnsd                                  0x102200ff9 after_resolve + 569
2   hnsd                                  0x1021fed04 hsk_cache_insert + 36
1   hnsd                                  0x102210939 hsk_dns_msg_encode + 41  dns.c:180
0   libsystem_malloc.dylib             0x7fff2021ffb7 _malloc_zone_malloc + 242
```

Repeating all this above on the pull request branch is nice and clean ;-)

pinheadmz (approved):
ACK e7cac4a2d9d4e8d638401cec1f2554ecd5fea925

```
Hash: SHA256

ACK e7cac4a2d9d4e8d638401cec1f2554ecd5fea925
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmGSp/cACgkQ5+KYS2KJ
yTpv0RAAtbBrRe1zsXIu3HlITHsbRmJi7LBuKhTL5AoEZMuylU5JQ7F02VhcWb+d
E24OKtpHFF3UzZGdd8rdO1htkCafdCDh6M3nVWni/9NKEoQt+URyxq5V99NZLvbu
LY2wmiIY1jbUhu7ZqimRqt5lSeVJzBeJkGaWgTH5MfTaOMXItnoOJLOPmxaecsRE
URw2M+pRlNjuDVzevBd60nOqo/aBycJBeTqnlfbiVSdyyhWlw8VInYrOaHrIbzf1
SONgO2+Yg+VaexQOFLagtwmoi0Ga+H7dnrEtjd/DkEyNO3LI25agBqRyS7AE9QYL
oE8ORuCNCja6nkIm64ZBAMHVINZIqr53J4WGJgY+I7LWto/yI5hLqQ8gQrpypzSc
ZbcY8hf+nEOXDI1FEcNmOQ8tFDfog/uCr1QekY8Yz5sPc5kLBNKLaZ/hJbBTQ1C4
eOErF9CBzGE5jZs7qojcmLM1YooetQOsV1bTZ80zPXltXug7IQD+6lgC0ZMDoEip
UeouvAfn2+N5auePfepkUU7Hwkf3S4YPiKujJvAyvcSlVUJpf6GFtLBioL0MjDHM
3TTcpYHJ+bdYE6TR4/LPfxsZKjVY01tgnYHilYmn3eVh+AvgZL8upX7a4d3bmEz7
9FMUgrnLRdnDU3sgwAPVpz97zzAcx0i6wxk3doCSzMe8bo1DMRE=
=wQtt
-----END PGP SIGNATURE-----
```

</details>

PR #81, opened 2021-12-15
daemon/pool: allow user to set supplemental user agent string

pinheadmz:
Usage:

```
./hnsd -a bacon

# or

./hnsd --user-agent bacon
```

Result (from hsd):

```
--> hsd-rpc getpeerinfo
[
  {
    "id": 13,
    "addr": "127.0.0.1:64061",
    "addrlocal": "127.0.0.1:10000",
    "services": "00000000",
    "servicenames": [],
    "relaytxes": false,
    "lastsend": 1639587864,
    "lastrecv": 1639587864,
    "bytessent": 35343,
    "bytesrecv": 249,
    "conntime": 1,
    "timeoffset": 0,
    "pingtime": -1,
    "minping": -1,
    "version": 1,
    "subver": "/hnsd:1.0.0/bacon",
    "inbound": true,
    "startingheight": 0,
    "besthash": "0000000000000000000000000000000000000000000000000000000000000000",
    "bestheight": -1,
    "banscore": 0,
    "inflight": [],
    "whitelisted": false
  }
]
```

### DIFF
--- a/src/daemon.c
+++ b/src/daemon.c
@@ -36,6 +36,7 @@ typedef struct hsk_options_s {
   uint8_t *identity_key;
   char *seeds;
   int pool_size;
+  char *user_agent;
 } hsk_options_t;
 
 static void
@@ -52,6 +53,7 @@ hsk_options_init(hsk_options_t *opt) {
   opt->identity_key = NULL;
   opt->seeds = NULL;
   opt->pool_size = HSK_POOL_SIZE;
+  opt->user_agent = NULL;
 }
 
 static void
@@ -141,6 +143,9 @@ help(int r) {
     "  -l, --log-file <filename>\n"
     "    Redirect output to a log file.\n"
     "\n"
+    "  -a, --user-agent <string>\n"
+    "    Add supplemental user agent string in p2p version message.\n"
+    "\n"
 #ifndef _WIN32
     "  -d, --daemon\n"
     "    Fork and background the process.\n"
@@ -156,9 +161,9 @@ help(int r) {
 
 static void
 parse_arg(int argc, char **argv, hsk_options_t *opt) {
-  const static char *optstring = "c:n:r:i:u:p:k:s:l:h"
+  const static char *optstring = "c:n:r:i:u:p:k:s:l:h:a"
 #ifndef _WIN32
-    "d"
+    ":d"
 #endif
     ;
 
@@ -172,6 +177,7 @@ parse_arg(int argc, char **argv, hsk_options_t *opt) {
     { "identity-key", required_argument, NULL, 'k' },
     { "seeds", required_argument, NULL, 's' },
     { "log-file", required_argument, NULL, 'l' },
+    { "user-agent", required_argument, NULL, 'a' },
 #ifndef _WIN32
     { "daemon", no_argument, NULL, 'd' },
 #endif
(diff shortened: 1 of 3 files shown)


src/daemon.c
```diff
@@ -156,9 +161,9 @@ help(int r) {
 
 static void
 parse_arg(int argc, char **argv, hsk_options_t *opt) {
-  const static char *optstring = "c:n:r:i:u:p:k:s:l:h"
+  const static char *optstring = "c:n:r:i:u:p:k:s:l:h:a"
 #ifndef _WIN32
-    "d"
+    ":d"
```
pinheadmz, line 166: this `:` is a bug fix, unless you ever tried using `-h` before you wouldn't know it was broken.

src/pool.c
```diff
@@ -173,6 +173,8 @@ hsk_pool_init(hsk_pool_t *pool, const uv_loop_t *loop) {
   pool->pending_count = 0;
   pool->block_time = 0;
   pool->getheaders_time = 0;
+  pool->user_agent = (char *)malloc(255);
```
buffrr, line 176: This should be freed in `hsk_pool_uninit` but if it's fixed size maybe it shouldn't use malloc here

src/pool.c
```diff
@@ -288,6 +290,25 @@ hsk_pool_set_seeds(hsk_pool_t *pool, const char *seeds) {
   return true;
 }
 
+bool
+hsk_pool_set_agent(hsk_pool_t *pool, const char *user_agent) {
+  assert(pool);
+
+  if (!user_agent)
+    return true;
```
buffrr, line 298: did you mean to return false instead?

pinheadmz: No, if the user does not supply an extra string, then this function does nothing which counts as success. The default user agent has already been applied.

src/pool.c
pinheadmz: I had trouble with this because its not a fixed size. There is a max size and the string starts out as `hsd:v0.0.1` but then more characters are added to that later. I think you're right about free though, will add and force push

pinheadmz: ok updated with `free`

buffrr: > I had trouble with this because its not a fixed size. There is a max size and the string starts out as hsd:v0.0.1

oh I see i meant something like `char user_agent[255]` defined in pool struct then something like

```C
strcpy(pool->user_agent, HSK_USER_AGENT);
strcat(pool->user_agent, "extra string ...");
```
but yeah this works too

pinheadmz: > oh I see i meant something like `char user_agent[255]` defined in pool struct then something like

this is one thing I tried but then I got `array type char can not be assigned` or something

pinheadmz: confirmed leak: 

```

leaks Report Version: 4.0
Process 59478: 171 nodes malloced for 77 KB
Process 59478: 1 leak for 256 total leaked bytes.

    1 (256 bytes) ROOT LEAK: 0x1091059b0 [256]  length: 12  "/hnsd:1.0.0/"
```

and confirmed no leak after adding `free()`

src/pool.c
buffrr: As a user of this library if i call `hsk_pool_set_agent(pool, NULL)` and it returned true i expect my supplemental agent to be cleared but can't imagine a use case for this :D so that's fine

pinheadmz: Oh I see huh. I wonder actually if passing `""` would work. It's an empty string but wuoldn't be a NULL pointer ?

buffrr: yup passing empty string clears the agent/returns true that's why NULL shouldn't indicate success here but i'm just being nitpicky it is not really important

buffrr (approved):
LGTM

PR #82, opened 2021-12-16
Daemon separate close and free

nodech:
### Fix malloc for user agent.
  https://github.com/handshake-org/hnsd/blob/d2bb5f4c77c2cb2aba18be7c57ce6ffe9abca821/src/pool.c#L309-L312
  strcat will put `\0` byte outside allocated memory when passed string with length `255-12`
  (12 -/hnsd:1.0.0/ length).

It did not crash, but still.
This will be command if you want to replicate issue:

```sh
./hnsd -a `node -e 'console.log("a".repeat(255 - 12))'` # Just run with aaaa...
gdb --args ./hnsd -a `node -e 'console.log("a".repeat(255 -12))'` # Gdb to check memory
```

### Separate close and free in the daemon

I noticed that daemon would segfault if the user agent check failed (`hsk_pool_set_agent`).
Reproduce with:
```sh
./hnsd -a `node -e 'console.log("a".repeat(255))'` 
```

The reason it fails is the fail label in the `hsk_daemon_init` that calls `hsk_daemon_after_close`.
`hsk_daemon_after_close`(this is basically close method for the main thread) does not differentiate between
stages of allocation and open and tries to close them anyway.

This PR introduces `hsk_daemon_uninit`, which only calls free for the daemon props. It is still called
by `hsk_daemon_after_close`, but the `hsk_daemon_init` will only call `uninit` function.

### DIFF
--- a/src/daemon.c
+++ b/src/daemon.c
@@ -399,6 +399,8 @@ static void
 hsk_daemon_after_close(void *data);
 static void
 hsk_daemon_signal_shutdown(void *data);
+static void
+hsk_daemon_uninit(hsk_daemon_t *data);
 
 int
 hsk_daemon_init(hsk_daemon_t *daemon, uv_loop_t *loop, hsk_options_t *opt) {
@@ -501,8 +503,34 @@ hsk_daemon_init(hsk_daemon_t *daemon, uv_loop_t *loop, hsk_options_t *opt) {
   return HSK_SUCCESS;
 
 fail:
-  hsk_daemon_after_close((void *)daemon);
+  hsk_daemon_uninit(daemon);
   return rc;
+}
+
+void
+hsk_daemon_uninit(hsk_daemon_t *daemon) {
+  if (!daemon)
+    return;
+
+  if (daemon->signals) {
+    hsk_signals_free(daemon->signals);
+    daemon->signals = NULL;
+  }
+
+  if (daemon->rs) {
+    hsk_rs_free(daemon->rs);
+    daemon->rs = NULL;
+  }
+
+  if (daemon->pool) {
+    hsk_pool_free(daemon->pool);
+    daemon->pool = NULL;
+  }
+
+  if (daemon->ns) {
+    hsk_ns_free(daemon->ns);
+    daemon->ns = NULL;
+  }
 }
 
 int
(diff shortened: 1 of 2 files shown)


pinheadmz:
Confirmed the misbehavior on daemon failing during init:

```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x59)
  * frame #0: 0x00000001000980e9 hnsd`uv_timer_stop(handle=0x0000000000000000) at timer.c:94:8 [opt]
    frame #1: 0x0000000100028e9a hnsd`hsk_pool_destroy [inlined] hsk_pool_close(pool=0x000000010a008200) at pool.c:370:7 [opt]
    frame #2: 0x0000000100028e8e hnsd`hsk_pool_destroy(pool=0x000000010a008200) at pool.c:384 [opt]
    frame #3: 0x00000001000023ef hnsd`hsk_daemon_init + 655
    frame #4: 0x0000000100002cb2 hnsd`main + 1602
    frame #5: 0x00007fff203f9f3d libdyld.dylib`start + 1
    frame #6: 0x00007fff203f9f3d libdyld.dylib`start + 1
```

So this is because we haven't actually started the UV loop yet? (uv gets started after `hsk_daemon_init()` returns). If we fail during init, we just need to uninit. If we fail after opening, we have more stuff to clean up.

pinheadmz (approved):
ACK 751bf4cf02c4fecb7bdb7836fec2c6d83a8a4a54

```
Hash: SHA256

ACK 751bf4cf02c4fecb7bdb7836fec2c6d83a8a4a54
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmG8wcUACgkQ5+KYS2KJ
yTqFaRAAqYk5vh0SPxvj3a1FZfZQaeAxIxqUer0LMph6nj2/xu7u/qW2CYJCmRG4
JrSnLoXDY8hE1v4jcUFrEzsM6yv0vXJr4PprolN/Y8OsOe7p4kVIoQDKVIf2MafN
o6mUoR5fTeQYULmZH9R4ol2TLr6hADEXRhjtI+tV/WbYeWnqkgEW52RLcnUor3XB
r+GJ9mLNFt1mCn16A02LtxrJgX6iU/bUpVnscgcycgC7fN0RyX4JTnOHool5toyx
Tsa4hrTAwbKf0Gq64k7YAymWlSZKFx3UEOBqHhog47+/1xOyizSGivES+J10hGIs
VQt6WrSx6IcKhY2vb4Bir0s3w2jBAvZDbULUtXfMZ85oe62LvQp8KUqZ3rui3ytW
5gTCnaXqMzpwtYE1pHg0QYdZ7k219T1pTzMinKFM+oNgDNN7cyt0ZGN3TQnP14mx
eSXLkDwNU6OO9rc9EvKjx4K0QIk0RViVnkORCY6wk3Hq5L0+hqNJjVPjUUGXMogh
b3oAGmtxyzT9x7ts6HK/AqHPxBamVE0m49E6GKWEDbbzGknUqwfVUq1QxohbQ8U7
Ntr2afwGbyCI31YsYyrINxeo38NtklAwgfgzDOBVjJAPpzlHClZ3l4QTxMrAXQAq
tM6APPs1Ii7csrvUg7fSfRN3Wlb2G2tYFjoJPW72amFfFlkIvK4=
=knn3
-----END PGP SIGNATURE-----
```

</details>

PR #89, opened 2022-01-31
dns: fix hsk_dns_label_from2 for max length label

pinheadmz:


test/hnsd-test.c
```diff
@@ -60,11 +60,54 @@ test_pointer_to_ip() {
...
+
+  hsk_dns_msg_t *msg1;
+  printf(" 62 char TLD\n");
+  char *name1 = "dddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd.";
+  msg1 = hsk_resource_to_dns(&res, name1, HSK_DNS_NS);
+  hsk_dns_rrs_t ns1 = msg1->ns;
+  hsk_dns_rr_t *rr1 = ns1.items[0];
+  hsk_dns_ns_rd_t *rd1 = rr1->rd;
+  assert(strcmp(rr1->name, name1) == 0);
+  assert(strcmp(rd1->ns, target) == 0);
```
buffrr, line 86: you should probably call `hsk_dns_msg_free(msg1)` at the end here to keep address sanitizer happy. Even though it's just a test, this helps us find issues/memory leaks that aren't necessarily in tests. Same thing for msg2

pinheadmz: good catch thanks, I forgot about the `malloc` in `hsk_resource_to_dns()`. Fixed and rebased, Tested with OSX `leaks` tool, all clear now:

```
leaks Report Version: 4.0
Process 4483: 209 nodes malloced for 16 KB
Process 4483: 0 leaks for 0 total leaked bytes.
```

PR #90, opened 2022-02-25
Refactor: use wire-formatted byte arrays to represent names internally, no more strings

pinheadmz:
description WIP


test/hnsd-test.c
```diff
@@ -27,21 +41,328 @@ test_base32() {
...
+        hsk_dns_ptr_rd_t *ptr = rr->rd;
+        hsk_dns_name_from_string(record_read_vector.name1, ptr->ptr);
+        break;
+      }
+      // case HSK_DNS_RP: {
+      //   hsk_dns_rp_rd_t *rp = rr->rd;
+      //   hsk_dns_name_from_string(record_read_vector.name1, rp->mbox);
+      //   hsk_dns_name_from_string(record_read_vector.name2, rp->txt);
+      //   break;
+      // }
```
pinheadmz, line 282: TODO, figured this out:

https://www.rfc-editor.org/rfc/rfc3597#section-4

>   To avoid such corruption, servers MUST NOT compress domain names
>    embedded in the RDATA of types that are class-specific or not well-
>    known.  This requirement was stated in [RFC1123] without defining the
>    term "well-known"; it is hereby specified that only the RR types
>    defined in [RFC1035] are to be considered "well-known".

RP is not considered "well known" since its not in 1035

src/dns.c
```diff
@@ -3351,7 +3196,7 @@ hsk_dns_rrs_clean(hsk_dns_rrs_t *rrs, uint16_t type) {
  */
 
 bool
-hsk_dns_is_subdomain(const char *parent, const char *child) {
+hsk_dns_is_subdomain(const uint8_t *parent, const uint8_t *child) {
```
nodech, line 3199: While playing with this method I wrote the tests for this and it fails when `child == parent`.

Test case can be included in `test/hnsd-test.c`. Third test will fail both on master and this branch.
So this bug is not related to the PR, but in general to this method.
```c
void
test_hsk_dns_is_subdomain() {
  printf("  test_hsk_dns_is_subdomain\n");

  typedef struct {
    uint8_t parent[255];
    uint8_t child[255];
    bool result;
  } hsk_dns_is_subdomain_t;

  const hsk_dns_is_subdomain_t tests[] = {
    {
      .parent = "\x09""handshake""\x03""org""\x00",
      .child = "\x04""hnsd""\x09""handshake""\x03""org""\x00",
      .result = true
    },
    {
      .parent = "\x04""hnsd""\x09""handshake""\x03""org""\x00",
      .child = "\x09""handshake""\x03""org""\x00",
      .result = false
    },
    {
      .parent = "\x09""handshake""\x03""org""\x00",
      .child = "\x09""handshake""\x03""org""\x00",
      .result = true
    },
    {
      .parent = "\x09""handshake""\x03""org""\x00",
      .child = "\x04""test""\x03""org""\x00",
      .result = false
    }
  };

  for (int i = 0; i < ARRAY_SIZE(tests); i++) {
    const hsk_dns_is_subdomain_t test = tests[i];
    bool result = hsk_dns_is_subdomain(test.parent, test.child);
    assert(result == test.result);
  }
}
```
Javascript version will return expected results.
https://github.com/chjj/bns/blob/1bc5d056f4372e62cefd43112ca8b1ebc3372f86/lib/util.js#L317-L319

src/dns.h
```diff
@@ -588,26 +605,74 @@ hsk_dns_txt_alloc(void);
...
+/**
+ * Returns: boolean (success)
+ * In:      name:  pointer to uncompressed wire-formatted name in a byte array
+ * Out:     data:  (optional) pointer to beginning of destination for compressed
+ *                 name in a DNS message serialized as a byte array
+ *          len:   length of data written (compressed length)
+ *          cmp:   (optional) map of labels and their position in the DNS
+ *                 message for label compression
+ */
+static bool
```
nodech, line 635: If the function is static, it's private to that file. So it should not be part of header.

src/ns.c
```diff
@@ -434,17 +443,21 @@ hsk_ns_onrecv(
...
-        || strcmp(req->tld, "tor") == 0 // OnioNS
-        || strcmp(req->tld, "zkey") == 0) { // GNS
+    if (memcmp(req->tld, "\x03""bit", 4) == 0 // Namecoin
+        || memcmp(req->tld, "\x03""eth", 4) == 0 // ENS
+        || memcmp(req->tld, "\x03""exit", 5) == 0 // Tor
+        || memcmp(req->tld, "\x03""gnu", 4) == 0 // GNUnet (GNS)
+        || memcmp(req->tld, "\x03""i2p", 4) == 0 // Invisible Internet Project
+        || memcmp(req->tld, "\x03""onion", 6) == 0 // Tor
+        || memcmp(req->tld, "\x03""tor", 4) == 0 // OnioNS
+        || memcmp(req->tld, "\x03""zkey", 5) == 0) { // GNS
```
nodech, line 460: Should not this instead be:
```
    if (memcmp(req->tld, "\x03""bit", 4) == 0 // Namecoin
        || memcmp(req->tld, "\x03""eth", 4) == 0 // ENS
        || memcmp(req->tld, "\x04""exit", 5) == 0 // Tor
        || memcmp(req->tld, "\x03""gnu", 4) == 0 // GNUnet (GNS)
        || memcmp(req->tld, "\x03""i2p", 4) == 0 // Invisible Internet Project
        || memcmp(req->tld, "\x05""onion", 6) == 0 // Tor
        || memcmp(req->tld, "\x03""tor", 4) == 0 // OnioNS
        || memcmp(req->tld, "\x04""zkey", 5) == 0) { // GNS
```
First byte size does not match the length of the label.

src/req.h
```diff
@@ -27,8 +27,11 @@ typedef struct {
   size_t max_size;
   bool dnssec;
 
+  // For Unbound
+  char namestr[HSK_DNS_MAX_NAME_STRING];
```
nodech, line 31: Instead of adding this here, I believe it's more suitable to have it allocated only in `hsk_rs_worker_resolve`, where it's actually needed. There's no need to carry full string version just for unbound call.

src/rs.c
```diff
@@ -377,9 +377,15 @@ hsk_rs_onrecv(
     goto fail;
   }
 
+  // Unbound's name resolution API expects a single null-terminated string.
+  // Since 0x00 is a valid byte mid-label in wire format we need to
+  // convert `req->name` to presentation format (i.e. "\000" for 0x00)
+  if (!hsk_dns_name_to_string(req->name, req->namestr))
```
nodech, line 383: > Instead of adding this here, I believe it's more suitable to have it allocated only in hsk_rs_worker_resolve, where it's actually needed. There's no need to carry full string version just for unbound call.
> from req.h comment

Here we can also get rid of this part and only translate to string inside hsk_rs_worker_resolve.

nodech (commented):
Instead of `char *` for the name we could use `typedef` and give better name. It will make code much more readable whenever we are working with names. This could also apply to `labels`, because labels are always in the form of `size + label`, it will clarify what to expect (also MAX_SIZE 63 + 1)

src/dns.h
```diff
@@ -617,125 +682,94 @@ hsk_dns_name_read_size(
...
  *          count:  number of labels in the name
  *          index:  index of starting label in name to return
  *                  (-1 = last label, TLD)
- * Out:     ret:    pointer to string to be filled with name from
- *                  specified label through end of name
+ * Out:     ret:    pointer to byte array to be filled with name from
+ *                  specified label through end of name including final 0x00
  */
 int
 hsk_dns_label_from2(
```
nodech, line 726: I think name of this method is misleading, it returns sub name - not a specific label.

src/dns.c
```diff
@@ -2379,145 +2495,47 @@ hsk_dns_name_read_size(
...
-) {
-  int size = hsk_dns_name_read_size(*data, *data_len, dmp);
-
-  if (size == -1)
-    return false;
-
-  char *n = malloc(size + 1);
-
-  if (!n)
+hsk_dns_name_verify(const uint8_t *name) {
```
nodech, line 2498: Name of this method in hnsd is misleading. It's `hsk_dns_tld_verify` instead of `name_verify`. In hsd it's easier to understand rules->nameVerify automatically means `tld` verify. But not in hnsd.

src/cache.c
```diff
@@ -256,15 +258,9 @@ hsk_cache_key_equal(const void *a, const void *b) {
 }
 
 bool
-hsk_cache_key_set(hsk_cache_key_t *ck, const char *name, uint16_t type) {
+hsk_cache_key_set(hsk_cache_key_t *ck, const uint8_t *name, uint16_t type) {
   assert(ck);
 
-  if (!hsk_dns_name_verify(name))
```
nodech, line 262: Don't we still need to verify if `tld` is correct ?

Otherwise this method will never return false.
I guess that could make this more precise and instead move tld check to the `hsk_cache_insert_data` and `hsk_cache_get_data` before the cache_key_set.

nodech: This method will become `void` and instead those two methods will return early.

src/cache.c
```diff
@@ -276,30 +272,6 @@ hsk_cache_key_set(hsk_cache_key_t *ck, const char *name, uint16_t type) {
     case 2:
       ref = !hsk_resource_is_ptr(name);
       break;
-    case 3:
```
nodech, line 279: As discussed with @pinheadmz, this should have been removed even before this update. As root wont parse/recognize SRV/TLSA//SMIMEA and OPENPGPKEY records at all.

src/resource.c
```diff
@@ -473,14 +466,22 @@ hsk_resource_has_ns(const hsk_resource_t *res) {
   return false;
 }
 
+void
+hsk_resource_write_synth(char *b32, uint8_t *name) {
```
nodech, line 470: Here we can just use `memcpy` instead of sprintfs, sprintf does not make it more readable or easier to work with.

src/resource.c
```diff
@@ -499,36 +500,33 @@ hsk_resource_to_ns(
...
         hsk_base32_encode_hex(c->inet6, 16, b32, false);
 
       // Magic pseudo-TLD can also be directly resolved by hnsd
-      sprintf(nsname, "_%s._synth.", b32);
+      hsk_resource_write_synth(b32, rd->ns);
     } else {
       // NS and GLUE records have the NS names ready to go.
-      assert(hsk_dns_name_is_fqdn(c->name));
-      strcpy(nsname, c->name);
+      memcpy(rd->ns, c->name, sizeof(c->name));
```
nodech, line 527: This sizeof will always be HSK_DNS_MAX_NAME.

src/resource.c
nodech: Also it's more readable and consistent with other methods, if the `uint8_t *name` - where we write - is first and `char *b32` - we copy from - second arg.

src/resource.c
```diff
@@ -1149,23 +1151,24 @@ hsk_resource_to_notimp(void) {
...
+  int len = hsk_dns_label_get(name, 0, label);
 
-  if (len < 2 || len > 29 || label[0] != '_')
+  // label is currently a wire-format serialized byte array,
+  // including the leading length byte.
+  if (len < 2 || len > 29 || label[1] != '_')
     return false;
 
-  int j = hsk_base32_decode_hex(&label[1], ip, false);
+  // We can cast label to a string for the base32 function
```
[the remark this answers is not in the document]
nodech: I see, so because it's base32 encoded(which should not contain 0x00).

src/req.c
```diff
@@ -83,32 +84,23 @@ hsk_dns_req_create(
...
+  memcpy(req->tld, &tld, tld[0] + 1);
 
   // Reference.
   req->ns = NULL;
 
   // DNS stuff.
   req->id = msg->id;
   req->labels = hsk_dns_label_count(qs->name);
-  strcpy(req->name, qs->name);
+  memcpy(req->name, qs->name, sizeof(qs->name));
```
nodech, line 103: sizeof will always be HSK_DNS_MAX_NAME.

src/pool.c
```diff
@@ -541,21 +544,21 @@ hsk_pool_resolve(
     return HSK_SUCCESS;
   }
 
-  hsk_name_req_t *head = hsk_map_get(&peer->names, req->hash);
```
nodech, line 544: `peer->names` is list of `tlds` instead, right ?

src/pool.c
```diff
@@ -1557,13 +1562,13 @@ hsk_peer_handle_proof(hsk_peer_t *peer, const hsk_proof_msg_t *msg) {
...
 
-  hsk_name_req_t *req, *next;
+  hsk_tld_req_t *req, *next;
 
   for (req = reqs; req; req = next) {
     next = req->next;
 
     req->callback(
-      req->name,
+      (uint8_t *)req->tld,
```
nodech, line 1571: here as well.

src/dns.c
```diff
@@ -594,7 +632,7 @@ hsk_dns_qs_init(hsk_dns_qs_t *qs) {
   assert(qs);
 
   memset(qs->name, 0, sizeof(qs->name));
```
nodech, line 634: sizeof will always be DNS_MAX_NAME

src/dns.c
```diff
@@ -1359,7 +1376,8 @@ hsk_dns_rd_write(
     case HSK_DNS_DNAME: {
       hsk_dns_dname_rd_t *r = (hsk_dns_dname_rd_t *)rd;
 
-      size += hsk_dns_name_write(r->target, data, cmp);
+      // no label compression
```
nodech, line 1379: So DNS_DNAME, DNS_SRV, DNS_RP, DNS_RRSIG and DNS_NSEC no longer use label compression?

src/dns.c
```diff
@@ -2098,87 +2120,100 @@ hsk_dns_name_parse(
...
-            b = 0xfe;
-
-          if (name)
-            name[noff] = b;
-
-          noff += 1;
-        }
+        // Label can not extend full name past maximum including length byte
+        // and resolved compression pointers.
+        if (noff + c + 1 > HSK_DNS_MAX_NAME)
```
nodech, line 2172: I think here is similar. This only ensures there's enough space for the rest of the label, but not the `0x00`.

This will make last test case we already have in `record_read_vectors.h`:
```c
  {
    HSK_DNS_CNAME,
    "123456789012345678901234567890123456789012345678901234567890123.123456789012345678901234567890123456789012345678901234567890123.123456789012345678901234567890123456789012345678901234567890123.12345678901234567890123456789012345678901234567890123456789012.",
    "",
    "\x3f""123456789012345678901234567890123456789012345678901234567890123\x3f""123456789012345678901234567890123456789012345678901234567890123\x3f""123456789012345678901234567890123456789012345678901234567890123\x3e""12345678901234567890123456789012345678901234567890123456789012\x00",
    256,
    false,
    "\x3f""123456789012345678901234567890123456789012345678901234567890123\x3f""123456789012345678901234567890123456789012345678901234567890123\x3f""123456789012345678901234567890123456789012345678901234567890123\x3e""12345678901234567890123456789012345678901234567890123456789012\x00",
    256
  }
```

To here earlier.

src/dns.c
```diff
@@ -2191,163 +2226,244 @@ hsk_dns_name_parse(
...
-      if (data) {
-        if (off + 1 + size > data_len) {
+        if (o < (2 << 13))
+          hsk_map_set(&cmp->map, &name[noff], (void *)o);
+      } else {
+        // If the string was already in the map,
+        // point to it instead of repeating the string.
+        // Ensure enough room for 2-byte pointer, this should only
+        // be an issue if we are pointing to a 1-byte label...!
+        if (2 + noff > HSK_DNS_MAX_NAME) {
```
nodech, line 2300: same here, `noff + 2`

src/dns.c
```diff
@@ -2191,163 +2226,244 @@ hsk_dns_name_parse(
...
-          data[off++] = ch;
-        }
+    if (label == 0x00) {
+      if (noff == 0) {
+        // Only one label and it's root
+        namestr[noff++] = '.';
+        namestr[noff] = '\0';
       } else {
-        off += size;
+        namestr[noff] = '\0';
```
nodech, line 2344: Do we not add last `.` ? `test.` or `test`, I am not sure what convention is that for names. (FQDN?)

src/dns.c
```diff
@@ -2191,163 +2226,244 @@ hsk_dns_name_parse(
...
+            continue;
+          }
+      }
 
-  if (i == 0 || name[i - 1] != '.') {
-    *len = off;
-    return false;
-  }
+      // Boring, printable character
+      namestr[noff++] = c;
```
nodech, line 2376: I think moving this inside `default` - after escaped char checks would make this more readable.

nodech (changes requested):
This change from cstring to wire seems to simplify a lot of things, so a good change. However there are couple of things that can be improved (Other than comments in the code).

1. Instead of using `uint8_t` which does not provide more information about the type we are working with, we could instead use typedef alias to it: e.g. `typedef uint8_t hsk_dns_name`, which will be used exactly as if it was uint8_t, but will simplify reading the code in a lot of places.
    - Side note: I was thinking of using `label` separately as well, but after review I think current label state is better. First of all it can be passed around to `name` related methods w/o problems, because it retains the name features and extra byte is worth the features. We could potentially have label alias as well, but in the end I think it's not worth the hassle, name is sufficient.
2.  memcpy usage can be improved a ton everywhere. Currently, when we work with full names that we want to move from one struct to another, we just do memcpy(..., MAX_SIZE), which is not optimal. Instead we could use custom `name_cpy` or something similar to achieve this. Especially, since most of our names will be far from reaching 255 bytes.  
  This could also allow for stack allocations to be more precise, instead of `uint8_t name[max] = {0}` for empty root, we could just use single byte on the stack: `uint8_t name[1] = {0}`. Of course this stack "optimization" does not apply to every case (e.g. when we don't know how big data we expect).
    - Here is an benchmark and one of the possible solutions implemented: https://github.com/nodech/hnsd-misc-c/tree/master/namecpy-v-memcpy
3. This is out of scope of this PR, but it would be nice to move out name related methods to `name.h`. Moving out constants to `dns-constants` can also help with this.
4. This tests also could use some organization, so upcoming PR: https://github.com/pinheadmz/hnsd/pull/6 can be checked here.

Okay, I think this is it for now. I am not sure I have not missed anything (caught couple things  today that I overlooked the other day.)

Great work !

src/resource.c
nodech: Here is the example of this method (not fully tested): https://github.com/nodech/hnsd-misc-c/tree/master/sprintf-v-memcpy

src/dns.c
pinheadmz: Confirmed that equal name should count here looking at knot resolver tests (and their `knot_dname_in_bailiwick()` function)

fixed and added test in fe6e1d5ea0439edc7fcd2d6e9668e3dd1c80ea9c

src/dns.h
pinheadmz: Yeah it returns a name "starting from a given label" i.e. "label from". I suppose we could rename it `hsk_dns_name_from_label()` ?

src/dns.h
```diff
@@ -6,8 +6,196 @@
 #include <stdlib.h>
 #include "map.h"
 
+// Constants
+// HSK_DNS_MAX_NAME includes all length bytes and final 0x00
+#define HSK_DNS_MAX_NAME 255
+// Byte arrays that include a length byte and final 0x00 (".")
+// will need to be size HSK_DNS_MAX_LABEL + 2
```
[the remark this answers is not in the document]
pinheadmz: > But because currently we don't have a way to differentiate, maybe it's easier that everything is in the name format 

This is my thinking as well. So all single labels or TLDs etc work just like full names do, i.e. they are read one label at a time starting with length until a label with length 0 is reached (the final 0x00)

src/pool.c
pinheadmz: yeah the names we have requested proofs for from a given peer.

src/dns.c
```diff
@@ -368,10 +404,10 @@ hsk_dns_msg_read(uint8_t **data, size_t *data_len, hsk_dns_msg_t *msg) {
     if (!qs)
       goto fail;
 
+    hsk_dns_rrs_push(&msg->qd, qs);
```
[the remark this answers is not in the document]
pinheadmz: Yes seems like it. I had to stare at this for a minute to understand why I did it, then realized I left us all an explanation in the commit message :-P

1aa326d682a3f8df90a6b652635a7773282e05cc

src/dns.c
```diff
@@ -431,13 +469,13 @@ hsk_dns_msg_read(uint8_t **data, size_t *data_len, hsk_dns_msg_t *msg) {
       msg->edns.rd = opt->rd;
       msg->code |= msg->edns.code << 4;
 
+      hsk_dns_rr_t *popped = hsk_dns_rrs_pop(&msg->ar);
+      assert(popped == rr);
       free(rr->rd);
```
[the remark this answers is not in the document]
pinheadmz: `hsk_dns_rr_free` would actually free too much and I tried it, and ended up with `pointer being freed was not allocated` thrown from `hsk_dns_msg_uninit(msg=0x0000000103009000) at dns.c:99:5` which is the big uninit at the very end of that object's life. The reason is because we still want to save some of the data from the original message:

```
      msg->edns.rd_len = opt->rd_len;
      msg->edns.rd = opt->rd;
```

If we `hsk_dns_rr_free()` this memory would be freed, and then we'll try to free that memory again later. The OPT record is a bit funny, especially the variable naming here, because it has its own record data or `rd` member.

So we have:

```
typedef struct hsk_dns_rr_s {
  uint8_t name[HSK_DNS_MAX_NAME];
  uint16_t type;
  uint16_t class;
  uint32_t ttl;
  void *rd; --->>> which for OPT will be hsk_dns_opt_rd_t...
} hsk_dns_rr_t;

typedef struct {
  size_t rd_len;
  uint8_t *rd; --->>> this data is copied to msg.edns.rd and freed by hsk_dns_msg_uninit()
} hsk_dns_opt_rd_t;
```


essentially `rr->rd->rd` 🙄

src/dns.c
pinheadmz: Correct! explanation in commit message: 5d846f9cdb1fe3425b34e8f32472a38e68ab80e8

src/dns.c
```diff
@@ -2098,87 +2120,100 @@ hsk_dns_name_parse(
...
         if (ptr == 0)
           res = off;
 
+        // Don't allow more than 10 pointers.
+        // I think this is a bit of a hack to prevent infinite loops.
+        // knot-dns simply requires that pointers always point backwards
+        // in the message stream. We can't do that with this function's
+        // inputs because we don't actually know where `data` appears in `dmp`.
         ptr += 1;
-
```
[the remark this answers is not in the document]
pinheadmz: eh, I think this is a TODO as described in my rudely critical comment. 10 pointers isn't any kind of DNS rule it really is just a hack and we can do better...

src/dns.c
pinheadmz: We do, on line 2381. Right here we're just adding the dot in the edge case that the entire name is just `"."` with no other labels

src/dns.c
pinheadmz: ok cool added in f5200ab1251950b897e8ef3f71f5a11737aeaa5a

src/dns.c
pinheadmz: thanks fixed in f5200ab1251950b897e8ef3f71f5a11737aeaa5a

src/dns.c
```diff
@@ -2191,163 +2226,244 @@ hsk_dns_name_parse(
...
+      // Check label map for remainder of name
+      size_t p = (size_t)hsk_map_get(&cmp->map, &name[noff]);
+
+      // Add remainder of name to map if it's not there already
+      if (p == 0) {
+        size_t o = (size_t)(&data[off] - cmp->msg);
 
-      if (data) {
-        if (off + 1 + size > data_len) {
+        if (o < (2 << 13))
```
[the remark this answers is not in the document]
pinheadmz: added in f5200ab1251950b897e8ef3f71f5a11737aeaa5a

src/dns.c
```diff
@@ -2191,163 +2226,244 @@ hsk_dns_name_parse(
...
+    }
+
+    // Label size can not exceed maximum
+    if (label > HSK_DNS_MAX_LABEL) {
+      *len = off;
+      return false;
+    }
+
+    // Label plus size byte can not extend full name past maximum
+    if (label + 1 + noff > HSK_DNS_MAX_NAME) {
```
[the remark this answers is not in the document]
pinheadmz: awesome thanks, fixed in f5200ab1251950b897e8ef3f71f5a11737aeaa5a

src/dns.c
```diff
@@ -2191,163 +2226,244 @@ hsk_dns_name_parse(
...
-  for (s = (char *)name, i = 0; *s; s++, i++) {
-    if (name[i] == '.') {
-      if (i > 0 && name[i - 1] == '.') {
+  for (;;) {
+    uint8_t label = name[noff];
+
+    // End of name "."
+    if (label == 0x00) {
+      // Almost made it! Last byte exceeds max
+      if (1 + noff > HSK_DNS_MAX_NAME) {
```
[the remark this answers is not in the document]
pinheadmz: sure, f5200ab1251950b897e8ef3f71f5a11737aeaa5a

src/dns.c
pinheadmz: fixed in f5200ab1251950b897e8ef3f71f5a11737aeaa5a

src/dns.c
```diff
@@ -2098,87 +2120,100 @@ hsk_dns_name_parse(
...
 
     switch (c & 0xc0) {
+      // Uncompressed label
       case 0x00: {
+        // Label size can not exceed maximum
         if (c > HSK_DNS_MAX_LABEL)
           return -1;
 
+        // Label size can not exceed length of data we are parsing
         if (off + c > data_len)
```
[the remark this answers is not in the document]
pinheadmz: great catch thanks, add the +2 and text vectors in f5200ab1251950b897e8ef3f71f5a11737aeaa5a

src/dns.c
pinheadmz: fixed in f5200ab1251950b897e8ef3f71f5a11737aeaa5a

src/dns.c
```diff
@@ -719,19 +756,10 @@ hsk_dns_rr_free(hsk_dns_rr_t *rr) {
 }
 
 bool
-hsk_dns_rr_set_name(hsk_dns_rr_t *rr, const char *name) {
+hsk_dns_rr_set_name(hsk_dns_rr_t *rr, const uint8_t *name) {
```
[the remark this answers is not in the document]
pinheadmz: fixed in f5200ab1251950b897e8ef3f71f5a11737aeaa5a

src/dns.c
```diff
@@ -108,7 +144,7 @@ hsk_dns_msg_write(const hsk_dns_msg_t *msg, uint8_t **data) {
 
   if (data) {
     cmp = &cmp_;
-    hsk_map_init_str_map(&cmp->map, NULL);
```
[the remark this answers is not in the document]
pinheadmz: nice, fixed in f5200ab1251950b897e8ef3f71f5a11737aeaa5a

src/ns.c
```diff
@@ -333,7 +333,9 @@ hsk_ns_onrecv(
   // The synth name then resolves to an A/AAAA record that is derived
   // by decoding the name itself (it does not have to be looked up).
   bool should_cache = true;
-  if (strcmp(req->tld, "_synth") == 0 && req->labels <= 2) {
+  if (memcmp(req->tld, "\x06""_synth", 7) == 0
```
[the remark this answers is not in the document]
pinheadmz: I don't think the last 0x00 matters really because we have the length byte already and `req->tld` wouldn't ever end with anything else. But, sure! added in f5200ab1251950b897e8ef3f71f5a11737aeaa5a

src/pool.c
pinheadmz: fixed in f5200ab1251950b897e8ef3f71f5a11737aeaa5a

src/pool.c
```diff
@@ -693,7 +696,7 @@ hsk_peer_timeout_reqs(hsk_peer_t *peer) {
       next = req->next;
 
       req->callback(
-        req->name,
+        (uint8_t *)req->tld,
```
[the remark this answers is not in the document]
pinheadmz: fixed in f5200ab1251950b897e8ef3f71f5a11737aeaa5a

src/pool.c
```diff
@@ -652,13 +655,13 @@ hsk_pool_merge_reqs(hsk_pool_t *pool, hsk_map_t *map) {
...
   if (pool->pending_count > 100) {
-    hsk_name_req_t *req, *next;
+    hsk_tld_req_t *req, *next;
 
     for (req = pool->pending; req; req = next) {
       next = req->next;
 
       req->callback(
-        req->name,
+        (uint8_t *)req->tld,
```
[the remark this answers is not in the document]
pinheadmz: :+1: thanks, fixed in f5200ab1251950b897e8ef3f71f5a11737aeaa5a

but actually looks like `hsk_pool_merge_reqs()` is never actually used either! (only reference is commented out)

src/req.c
pinheadmz: :+1: fixed in f5200ab1251950b897e8ef3f71f5a11737aeaa5a

src/resource.c
pinheadmz: > Also it's more readable and consistent with other methods, if the `uint8_t *name` - where we write - is first and `char *b32` - we copy from - second arg.

I didn't make this change yet because I think the hnsd style for these kinds of methods is more like `_write(in, out)`, such as `hsk_dns_msg_write(const hsk_dns_msg_t *msg, uint8_t **data)`. Even though I think you're right about standard methods like `memcpy(dest, src)` ? 

I copied your memcpy implementation of this function from your benchmark test, thanks! 

fixed in f5200ab1251950b897e8ef3f71f5a11737aeaa5a

src/cache.c
pinheadmz: Good point, I think we can remove the return value. We check `hsk_dns_name_verify()` in `hsk_ns_onrecv()` which is the entry point for the user's request and is checked before attempting name resolution and before caching. So I think throughout hnsd we can assume that `req->tld` is already valid.

so I changed `hsk_cache_key_set()` to void -- what did you mean by "two methods" ?

fixed in f5200ab1251950b897e8ef3f71f5a11737aeaa5a

src/ns.c
pinheadmz: :+1: fixed in f5200ab1251950b897e8ef3f71f5a11737aeaa5a

pinheadmz:
> * Here is an benchmark and one of the possible solutions implemented: https://github.com/nodech/hnsd-misc-c/tree/master/namecpy-v-memcpy

Not sure why but when I compiled and ran this on my computer, the `just_cpy` method was way faster:

```
just_cpy: ops=10000, time=0.003070s, rate=3257328 ops/s 307 ns/op
name_cpy: ops=10000, time=0.004199s, rate=2381519 ops/s 419 ns/op
```

with `-O2`
```
just_cpy: ops=10000, time=0.000001s, rate=10000000000 ops/s 0 ns/op
name_cpy: ops=10000, time=0.000932s, rate=10729613 ops/s 93 ns/op
```

src/dns.c
```diff
@@ -2191,163 +2226,244 @@ hsk_dns_name_parse(
...
+          return false;
 
-  if (data) {
-    if (off >= data_len) {
-      *len = off;
-      return false;
+        if (label + 1 > HSK_DNS_MAX_LABEL)
+          return false;
+
+        // Only a single character has been escaped, write it and advance
```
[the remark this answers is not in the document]
pinheadmz: I don't think this is necessary. Knot resolver doesn't judge the character that follows the escape slash:

https://gitlab.nic.cz/knot/knot-dns/blob/master/src/libknot/dname.c#L405-435

and neither does bind apparently:

```
--> dig \g\o\o\g\l\e\.\c\o\m +short
142.251.40.110
```

src/dns.c
pinheadmz: > It's worth to check if this needs to be returned back to the methods we removed this as well as check dirty names. Even though dirty names are not a thing, we still need to make sure TLDs are following Handshake rules on top of DNS ones.

Again I think that since we are sanitizing all input with this in `hsk_ns_onrecv()` I don't think we need to put it back anywhere else downstream ...?

src/dns.c
```diff
@@ -2527,119 +2545,123 @@ hsk_dns_name_cmp(const char *a, const char *b) {
...
 
-  if (alen < blen)
-    return -1;
+  return 0;
+}
 
-  if (alen > blen)
-    return 1;
+bool
+hsk_dns_name_equal(const void *a, const void *b) {
```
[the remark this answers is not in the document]
pinheadmz: I think it's because we pass this function to `hsk_map_init_map()` which accepts as an argument one of these:

```
typedef bool (*hsk_map_equal_func)(const void *a, const void *b);
```

src/ns.c
```diff
@@ -770,13 +786,13 @@ after_resolve(
...
 static int
-hsk_tld_index(const char *name) {
+hsk_tld_index(const uint8_t *tld) {
   int start = 0;
   int end = HSK_TLD_SIZE - 1;
 
   while (start <= end) {
     int pos = (start + end) >> 1;
-    int cmp = strcasecmp(HSK_TLD_NAMES[pos], name);
+    int cmp = strcmp(HSK_TLD_NAMES[pos], (char *)&tld[1]);
```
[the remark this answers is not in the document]
pinheadmz: Updated `tld.h` generated from this branch of hs-names: https://github.com/handshake-org/hs-names/pull/12

Also had to fix our name compare function to handle labels of different lengths correctly. Everything updated and fixed in b798359d025cfe57d22104bb642f1f99271a22a3

src/req.h
pinheadmz: great call thanks, added in 1fffc2b831cdd53bc14c46d18122556f54848912

src/rs.c
pinheadmz: done in 1fffc2b831cdd53bc14c46d18122556f54848912

PR #92, opened 2022-03-10
Remove config.h.in

nodech:
`config.h.in` is generated file and it may differ from system to system, appearing in the modified list.
`.gitignore` already contains that file, it was probably committed before the .gitignore update.


buffrr:
Yeah it shouldn't really be committed to the repo. Hm build fails when removed (CI builds show this too).

```
configure.ac:22: error: required file 'src/config.h.in' not found
Makefile.am: installing './depcomp'
Error: Process completed with exit code 1.
```

> .gitignore already contains that file, it was probably committed before the .gitignore update.

The .gitignore in the root directory doesn't have `src/config.h.in` only `config.h.in` so that's probably why it was included I think.


It seems that `autogen.sh` is not generating the file correctly. I'm not sure why this script doesn't just call `autoreconf` which does all that (see [bitcoin core](https://github.com/bitcoin/bitcoin/blob/master/autogen.sh) as an example): 

```diff
-"$LIBTOOLIZE" --copy
-"$ACLOCAL" -I m4
-"$AUTOCONF"
-"$AUTOMAKE" --add-missing --copy
+"$AUTORECONF" --install
```

nodech:
oh, `autogen.sh` fails when you have fresh clone (no autogened files). Easier to reproduce :)

nodech:
We can either run `autoheader` in the `autogen.sh` which can generate `config.h.in` file or commit new header files generated by updated tools.

I am leaning towards just adding autoheader step before automake.

buffrr (approved):
LGTM

PR #98, opened 2022-09-29
Initialize from checkpoint

pinheadmz:
Why start syncing from genesis if we don't have to? Block 135000 is a few months old and hopefully won't ever change. Let's start *from there!*

opt-in to faster sync with `-t` or `--checkpoint`.

Includes bash script used to generate `checkpoint.h` which stores the hex encoded headers for 150 blocks starting at any given height (we need this many headers so hnsd can accurately compute target difficulty in all "new" blocks that follow in the chain). It also stores the current accumulated chainwork which is needed to insert the first hard-coded header (#135000).

When active, on startup after pool and chain initialization, `daemon` injects the set of headers into `chain` and sync begins with block #135150.

Thinking ahead: we can use this mechanism to store (and prune) headers on disk for hnsd. When we reboot, we only need to read 150 headers from disk to continue syncing. If HNS ever encounters a chain reorganization deeper than 144 blocks (!) hnsd should naturally fall back to the genesis block and attempt to resync from there. (This would be a result of sending an otherwise empty chain locator to peers to request blocks)


pinheadmz:
Moving to https://github.com/handshake-org/hnsd/pull/99

PR #100, opened 2022-10-06
update openwrt Makefile

shalzz:
This disables compiling with -O2 flag since it breaks cross-compiling with at least the MIPS architecture.

Optimizations should be safe to re-enable when the build is fixed to handle COPTS compile flags correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,12 +286,12 @@ $ docker stop hnsd
 ### OpenWRT
 
 To build hnsd as an OpenWRT package you'll need to rename `openwrt_Makefile` to `Makefile` 
-and put it in `your_openwrt_dir/package/net/hnsd` before building.
+and put it in `your_openwrt_dir/feeds/package/net/hnsd` before building.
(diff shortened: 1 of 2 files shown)


pinheadmz:
Awesome, thanks so much for this contribution. If possible, can you explain how I could reproduce the issue? (like, do you own a MIPS device or something?). It would be nice of course to keep optimizations even if it means more work to fix the build.

shalzz:
> Awesome, thanks so much for this contribution. If possible, can you explain how I could reproduce the issue? (like, do you own a MIPS device or something?). It would be nice of course to keep optimizations even if it means more work to fix the build.

Yes, I'm build this for a MIPS device when I encounter this issue. 
You can reproduce by having at least these in your `.config` file when building openwrt.
```
CONFIG_TARGET_ath79=y
CONFIG_TARGET_ath79_generic=y
CONFIG_TARGET_ath79_generic_DEVICE_tplink_archer-a7-v5=y
```

PR #101, opened 2022-10-14
Organize unit tests and add integration tests with hsd

### LINKED ISSUE #23: Test Framework
It would be ideal to add unit tests to `hnsd` so that its easier to develop without regressions. There are many options for test frameworks in C, many projects opt to roll their own framework. I found what I believe to be good options.

### Options

- [cmocka](https://cmocka.org/)
  - Used by bind, libssh
  - [Custom assertions](https://api.cmocka.org/group__cmocka__asserts.html)
  - Simple to use, create an array of `CMUnitTest`s using `cmocka_unit_test` which wrap `describe` like functions

- [check](https://libcheck.github.io/check/doc/check_html/index.html#Top)
  - Designed to work with autotools
  - Simple, but a lot of boilerplate

- [CGreen](https://cgreen-devs.github.io/index.html#_cgreen_quickstart_guide)
  - Large, has many features
  - TDD/BDD support, along with `Before`, `After`, etc macros

- Custom
  - No extra dependencies
  - As simple as it needs to be
  - Must implement it, must be bugfree

----

I'd personally opt for using `cmocka` since it looks simple to use and is also used by other applications that are similar (DNS + security). It would be easier than creating our own custom test library.

pinheadmz:
Closes #23 

This cleans up the framework for unit tests (written C and compiled into executable `test_hnsd`) and also introduces a new integration test system using nodejs and hsd.

```

--> ./test_hnsd
Testing hnsd...
test_base32
 test_base32_encode_decode
test_resource
 test_resource_pointer_to_ip
ok
--> cd integration/
--> npm run test

> hnsd_integration_test@1.0.0 test
> bmocha --reporter spec test/*.js


  Basic sync & resolve
    ✓ should register names and sync (845ms)
    ✓ should resolve DS
    ✓ should resolve NS
    ✓ should resolve TXT
    ✓ should resolve GLUE4 without glue
    ✓ should resolve GLUE4 with glue
    ✓ should resolve GLUE6 without glue
    ✓ should resolve GLUE6 with glue
    ✓ should resolve SYNTH4 with glue
    ✓ should resolve SYNTH6 with glue

  10 passing (953ms)

```

### DIFF
--- a/integration/package.json
+++ b/integration/package.json
@@ -1,0 +1,12 @@
+{
+  "name": "hnsd_integration_test",
+  "scripts": {
+    "test": "bmocha --reporter spec test/*.js",
+    "test-file": "bmocha --reporter spec"
+  },
+  "dependencies": {
+    "bmocha": "^2.1.5",
+    "bsert": "0.0.10",
+    "hsd": "^4.0.1"
+  }
+}
--- a/integration/package.json
+++ b/integration/package.json
@@ -1,5 +1,7 @@
 {
   "name": "hnsd_integration_test",
+  "meta": "!! version must match hnsd version in Makefile.am !!",
+  "version": "1.0.0", 
   "scripts": {
     "test": "bmocha --reporter spec test/*.js",
     "test-file": "bmocha --reporter spec"
--- a/integration/test-util.js
+++ b/integration/test-util.js
@@ -1,16 +1,21 @@
 
 'use strict';
 
-const {spawn} = require('child_process');
+const {spawn, execSync} = require('child_process');
+const assert = require('bsert');
 const path = require('path');
 const {FullNode} = require('hsd');
 const StubResolver = require('bns/lib/resolver/stub');
+const {EOL} = require('os');
+const {version} = require('./package.json');
+const network = 'regtest';
+const hnsdPath = path.join(__dirname, '..', 'hnsd');
 
 class TestUtil {
   constructor() {
     this.node = new FullNode({
       memory: true,
-      network: 'regtest',
+      network,
       listen: true,
       port: 10000,
       noDns: true,
@@ -27,13 +32,20 @@ class TestUtil {
   }
 
   async open() {
+    const hnsdVersion = (await execSync(hnsdPath + ' -v')).toString('ascii');
+    assert.strictEqual(
+      hnsdVersion,
+      `${version} (${network})` + EOL,
+      'Network or version mismatch'
+    );
+
     await this.node.open();
     await this.node.connect();
 
     this.wallet = this.node.plugins.walletdb;
 
     this.hnsd = spawn(
-      path.join(__dirname, '..', 'hnsd'),
+      hnsdPath,
       ['-s', '127.0.0.1:10000']
     );
 
--- a/integration/test-util.js
+++ b/integration/test-util.js
@@ -4,7 +4,7 @@
 const {spawn} = require('child_process');
 const path = require('path');
 const {FullNode} = require('hsd');
-const dns = require('bns/lib/dns');
+const StubResolver = require('bns/lib/resolver/stub');
 
 class TestUtil {
   constructor() {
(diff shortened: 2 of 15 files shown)


integration/test-util.js
```diff
@@ -0,0 +1,110 @@
...
+    // // TODO: Fix hnsd stdout parsing for chain height
+    //   setTimeout(() => {
+    //     reject(new Error('Timeout waiting for sync'));
+    //   }, 5000);
+    //   setInterval(() => {
+    //     if (this.hnsdHeight === this.node.chain.height)
+    //       resolve();
+    //   }, 100);
+    });
+  }
```
Falci, line 107:
```suggestion
  async waitForSync() {
    if (!this.hnsd || !this.hnsd.killed) {
      return new Promise(resolve => setTimeout(resolve, 100))
        .then(() => this.waitForSync);
    }
  }
```

integration/test-util.js
```diff
@@ -0,0 +1,110 @@
...
+
+      for (const line of lines) {
+        const words = line.split(/\s+/);
+
+        if (words[0] !== 'chain' || words.length < 2)
+          continue;
+
+        this.hnsdHeight = parseInt(words[1].slice(1, -2));
+      }
+    });
```
Falci, line 53:
```suggestion    let timer;
    this.hnsd.stdout.on('data', (data) => {
      const chunk = data.toString('ascii');

      // check for "chain (xxx)"
      const height = chunk.match(/chain \((?<height>\d+)\).*(?![\s\S]*chain)/)?.groups?.height;
      
      if (height) 
        this.hnsdHeight = parseInt(height);

      if (timer) 
        clearTimeout(timer);

      if (!this.hnsd.killed) 
        timer = setTimeout(() => {
          this.hnsd.kill();
        }, 200);
    });
```

Falci (changes requested):

pinheadmz:
Update: added command `make e2e` in makefile so integration tests can be launched easily from root directory

integration/package.json
```diff
@@ -0,0 +1,12 @@
...
+  "name": "hnsd_integration_test",
+  "scripts": {
+    "test": "bmocha --reporter spec test/*.js",
+    "test-file": "bmocha --reporter spec"
+  },
+  "dependencies": {
+    "bmocha": "^2.1.5",
+    "bsert": "0.0.10",
+    "hsd": "^4.0.1"
+  }
```
rithvikvibhu, line 11: I know `bns` is installed because of hsd, but since the test script explicitly uses `bns`, should be be included here?

rithvikvibhu (approved):
Very cool CI with 3 platforms! Both hnsd and integration tests work great locally.
I did forget to build for regtest (like the readme mentions) and the tests just timed out with no specific errors.

Can hnsd print the network name on start? It would help identify binaries and can also make the test suite confirm that it's build with the correct network. Just an idea, not blocking this PR.

pinheadmz:
> Can hnsd print the network name on start? It would help identify binaries and can also make the test suite confirm that it's build with the correct network. Just an idea, not blocking this PR.

Yeah re-building for each network is annoying, maybe we can work that out in the future and start hnsd with a network argument. I think we need to avoid using hnsd's stdout for anything important, and using HS class won't really work either since the "API" ports will be different for each network.

I think the fastest solution to this may be a version command or something that could work like this:

```
$ hnsd --version
1.0.0 (regtest)
```

pinheadmz:
@rithvikvibhu added in 1715b09d04a2629c8c6a03cad54d48a0bbeff364


```
--> make e2e
npm --prefix ./integration/ install

up to date, audited 37 packages in 332ms

found 0 vulnerabilities
npm --prefix ./integration/ run test

> hnsd_integration_test@1.0.0 test
> bmocha --reporter spec test/*.js


  Basic sync & resolve
    1) "before all" hook

  0 passing (89ms)
  1 failing

  1) Basic sync & resolve
       "before all" hook:

      AssertionError [ERR_ASSERTION]: Network or version mismatch

      Expected values to be strictly equal:
      
      + actual - expected
      
      + '1.0.0 (main)\n'
      - '1.0.0 (regtest)\n'
                ^
      
      assert.strictEqual(
             ^

      at TestUtil.open (/Users/matthewzipkin/Desktop/work/hnsd/integration/test-util.js:36:12)
      at async Context.<anonymous> (/Users/matthewzipkin/Desktop/work/hnsd/integration/test/sync-resolve-test.js:18:5)

make: *** [e2e] Error 1
```

rithvikvibhu:
Sweet, version shows network and test fails. LGTM

PR #103, opened 2022-10-23
chain: fix hsk_chain_maybe_sync

buffrr:
This is a small PR to fix `hsk_chain_maybe_sync`. When this function is called the first time, it sets `chain->synced = true`. So all sync checking logic afterwards gets skipped:

https://github.com/handshake-org/hnsd/blob/cf19505f3c0b473e7da800165e42fca36c649019/src/chain.c#L265-L269

If you were wondering why you see this in the logs it's because of that ^. 

```
chain (0): chain is fully synced
```

I think this was just old code before mainnet launch. After removing it, hnsd will refuse to answer requests until it's fully synced which is neat! Answering queries without waiting to sync is problematic due to cache poisoning from old records ... etc. Fingertip mitigates that by trying to guess the synced state but hnsd should ideally do it.

```
pool: sending proof request for: proofofconcept.
pool: cannot send proof request: chain not synced.
```

### DIFF
--- a/src/chain.c
+++ b/src/chain.c
@@ -262,12 +262,6 @@ hsk_chain_maybe_sync(hsk_chain_t *chain) {
 
   int64_t now = hsk_timedata_now(chain->td);
 
-  if (now < HSK_LAUNCH_DATE) {
-    hsk_chain_log(chain, "chain is fully synced\n");
-    chain->synced = true;
-    return;
-  }
-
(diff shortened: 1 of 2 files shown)


pinheadmz:
cherry-picked this into https://github.com/handshake-org/hnsd/pull/102

We can merge it there as part of the HS interface. 

THANKS BRUH!

Issue #96, opened 2022-08-11
Suffix user agent with slash


rithvikvibhu:
Not a big deal, but hnsd user agents don't end with a slash. This was added in hnsd (https://github.com/handshake-org/hnsd/pull/81) last year (before hsd https://github.com/handshake-org/hsd/pull/710 a few months ago) and hence the difference.

Taken from a node peer list:
```json
{
  "/hnsd:1.0.0/": 474,
  "/hnsd:1.0.0/beacon": 2,
  "/hsd:4.0.0/": 4,
  "/hsd:4.0.1/bob-wallet:1.0.0/": 1,
  "/hsd:4.0.2/": 7,
}
```

I think this needs to change?
https://github.com/handshake-org/hnsd/blob/856ee22e6755cf08eb8cc790cf1cd2822408ad5c/src/pool.c#L305-L312

PR #104, opened 2022-10-27
Stricter rules for custom user agent

pinheadmz:
Closes https://github.com/handshake-org/hnsd/issues/96

Blocked by #101 and #102 and #99 😬 

```
--> ./hnsd -a bad/agent
chain (0): chain is fully synced
addrman: added addr: 127.0.0.1:46806
failed adding user agent
failed initializing daemon: EFAILURE

--> ./hnsd -a good_agent
...
[info] (net) Received version (127.0.0.1:60772): version=1 height=0 services=0 agent=/hnsd:1.0.0/good_agent/
```


integration/test/config-test.js
```diff
@@ -0,0 +1,66 @@
...
+
+  before(async () => {
+    await util.open();
+  });
+
+  after(async () => {
+    await util.close();
+  });
+
+  describe.only('--agent', function () {
```
rithvikvibhu, line 25: `.only` to be removed? All tests pass without this.

integration/test/config-test.js
```diff
@@ -0,0 +1,66 @@
...
+    it('should have default user agent', async () => {
+      await util.generate(1);
+      await util.waitForSync();
+
+      const peer = util.node.pool.peers.head();
+      assert.strictEqual(peer.agent, '/hnsd:1.0.0/');
```
rithvikvibhu, line 31: Does the test need to be updated with every version bump? Could pick from package.json (which I think was the idea for adding version there)

Edit: yeah just saw version used below

rithvikvibhu (commented):
Agent commit looks good, tests run fine! Left some comments.

integration/test/config-test.js
pinheadmz: yep thanks

integration/test/config-test.js
pinheadmz: yep! fixing...

PR #130, opened 2025-03-18
Better Developer UX

connoralydon:
These code blocks add friction when first getting setup in a repo. The blocks are already defined as `sh` or `bash`, it is inconvenient to add $ in the block. Not everyone's prompt ends with a $.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ the source repo.
 ### OSX
 
 ``` sh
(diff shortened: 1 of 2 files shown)


Nathanwoodburn:
G'day,
These code blocks follow the standard of using `$` to indicate that the commands are to be run from a user (non-root) terminal.

PR #131, opened 2026-01-11
Update README.md

walletdomain:
git:// replaced with https://

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ DLLs to the hnsd directory, etc.
 ## Cloning
 
 ``` sh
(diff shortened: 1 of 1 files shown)


Nathanwoodburn:
Did you want to add the SSH url also?
`git@github.com:handshake-org/hnsd.git`

walletdomain:
If you think it would add value, then yes. My only focus was getting it out of a state of non-functioning.

Nathanwoodburn:
I figure might as well do both while you are submitting a PR

PR #121, opened 2023-11-24
Add workflow for updating tld.h file

Falci:
This PR adds a GitHub action script that will:
* Checkout and install [hs-names](https://github.com/handshake-org/hs-names) with its dependencies
* Download the TLD and zone file from ICANN
* Update the `src/tld.h`
* Commit and create a PR

There's no automatic trigger for this action. It needs to be manually triggered from the [actions page](https://github.com/handshake-org/hnsd/actions/workflows/update.yml).

PR #132, opened 2026-01-28
Update tld.h to fix issues with .in resolving

Nathanwoodburn:
The india TLD `.in` is having issues resolving for a lot of people. I've just updated the tld.h file using hs-names (see #121 for  how to do it yourself). This seems to have fixed the `.in` resolving so this PR is just to update the tld.h file for other people.

Resolving .in domains using the new tld.h file
image


Nathanwoodburn:
If possible, can we get this merged relatively quickly as it fixes resolution failures of a few TLDs (mainly .in). Thanks

PR #133, opened 2026-03-11
ci: Lock npm version and add brew prefix

Nathanwoodburn:
Fix actions to stop the consistent failures. Bumps hsd to the latest version to stop compilation errors on node versions newer than 20

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,14 +18,16 @@ jobs:
 
       - name: MacOS Install Deps
         if: matrix.os == 'macos-latest'
-        run: brew install automake autoconf libtool unbound
+        run: |
+          brew install automake autoconf libtool unbound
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -18,23 +18,25 @@ jobs:
...
       - name: Unit Tests
         run: ./test_hnsd
 
       - name: Setup Integration
         uses: actions/setup-node@v4
         with:
-          node-version: lts/*
-          check-latest: true
+          node-version: 20
+          check-latest: false
```
Falci: This can work as a temporary fix.
Maybe we should create an issue to revisit this.

Falci (approved):

.github/workflows/test.yml
Nathanwoodburn: This should be fixed now by increasing the HSD version in the integration testing https://github.com/Nathanwoodburn/hnsd/blob/cb1ab833ff8329d42a8d68e44e7c136196381d40/integration/package.json#L12

Falci (approved):

